PR #2, opened 2018-06-13
Serverless setup + README

bobheadxi:
- Scaffold serverless setup, mostly imported from https://github.com/nwhacks/hackpost, including the env setup.
- ~Remove licence - it was MIT in the original template by @chamkank and kept during my rebase, but on second thought its probably too permissive for our needs. Should discuss at the next all-hands if we want to pu a licence at all~
    - added back dual MIT-apache licence from 2018
- Added README(s) - project README has basic getting started instructions

todo:
- [x] look into google's https://serverless.com/framework/docs/providers/google/guide/credentials/ as mentioned in the ticket

Please dont merge when you review @chamkank


bobheadxi:
todo:
- [ ] look into google's https://serverless.com/framework/docs/providers/google/guide/credentials/

d4l3k:
MIT is too permissive? You worried about someone stealing your code? Pretty much all of the existing nwHacks code is MIT licensed.

d4l3k:
Also, are you planning on using firestore from the serverless functions only? Seems easier to just make calls directly from the front end js.

bobheadxi:
Hi @d4l3k!

> MIT is too permissive? You worried about someone stealing your code? Pretty much all of the existing nwHacks code is MIT licensed.

My thought was that MIT would cover any design assets (logos, etc) in the repository as well, and I wasn't sure how that's meant to be handled. I hadn't looked too closely at past website repos, but just noticed that you guys did put licences in there - I'll just copy those over tonight. Didn't see them on the first pass because they didn't show up on GitHub's licence thing 😛 

> Also, are you planning on using firestore from the serverless functions only? Seems easier to just make calls directly from the front end js.

We haven't discussed this for this year yet, but I assume @chamkank will make a decision on front at some point, though I did notice that's how it's been done in the past. Firebase dependencies are currently included in the web app, but we haven't actually started anything yet. I'm guessing you saw this:

```js
const integrations = {
  github: require('./integrations/github'),
  firestore: require('./integrations/firestore'),
}
```

thats a carryover from the other repo, forgot about it - i'll remove those for the time being as well. thanks for pointing this out!

d4l3k:
Ah, got it. Pretty sure the policy has just been to MIT everything and not worry about it.

d4l3k:
Everything else looks reasonable! Though that 5k line long yarn.lockfile is kind of WTF, but that's autogenerated so whatever.

bobheadxi:
I used to ignore that stuff (package-lock.json etc), but it's [recommended that they be checked in](https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control) so... oh well idk 🤷‍♂️

d4l3k:
Yeah, I realize

chamkank:
@d4l3k Thanks for your input & review Tristan! 😊

@bobheadxi 

> We haven't discussed this for this year yet, but I assume @chamkank will make a decision on that front at some point

As I mentioned in the Notion doc about the tech stack, and the `invite admin functionality` task on the Feature Board, we'll be using Firestore from the front end and serverless functions for everything else like emails.

chamkank:
@bobheadxi On that note I noticed you added Ava as the test framework. The plan was to use Jest for testing the React components (also on the Notion doc), and I think it would be nice if we used the same testing framework for everything.

serverless/serverless.yml
```diff
@@ -0,0 +1,25 @@
+service: hackpost
```
mingyokim: hackpost? should this be nwhacks2019?

bobheadxi: good catch! ill update this

bobheadxi:
Good point @chamkank, I've switched over to Jest. The dual MIT-Apache licence has also been added back with updated names + dates. I've also added some linting rules and added it as part of the Travis pipeline, and switched the web app compile to `docs/`

bobheadxi:
Google cloud functions isn't supported by serverless-offline, which so far seems to be the only solution of its kind (deploy serverless functions locally as if it was a traditional server to test with a web app). 

I found some alternatives, but they are all seem to be on a function-by-function basis and I either haven't been able to get them to work or they just don't do the same thing:

1. invoke functions individually by `serverless invoke local --function some_function`, deploy to cloud and hope it works
2. https://cloud.google.com/functions/docs/emulator - this is also for individual functions, and for me this just breaks

I will chuck both of these things on Notion, but I think we should meet to talk about what we want to do on that front. I feel like there's got to be a way to develop this stuff locally but idk, haven't found anything as nice as `serverless-offline` (AWS-only - I'm not even sure why it's AWS-only at this point but oh well)

So I added the google cloudfunction config to serverless.yml and removed the serverless-offline stuff, don't think I'll be adding anything else to this PR. @chamkank @mingyokim

chamkank:
I tried the GCF emulator, didn't work for me either. Let's talk about testing the serverless functions during Tuesday's meeting. [Deploying the function and testing through the console is also another option.](https://cloud.google.com/functions/docs/quickstart-console#test_the_function
)

(Side note: Added Windows support for `make web`)

bobheadxi:
sounds good, thanks all 🏃

PR #4, opened 2018-06-25
basic login functionality

mingyokim:
Basic user login page and login confirmation page.

Used https://github.com/firebase/firebaseui-web-react.

You can checkout and test locally with your own gmail account or test@example.com with pw=testtest.

This PR is built on top of #3, which initializes the firebase module.


web/components/login/SignedIn.js
```diff
@@ -0,0 +1,37 @@
...
+export default class SignedIn extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      user: null,
+    };
+  }
+
+  componentWillMount() {
+    this.removeAuthListener = firebase.auth().onAuthStateChanged((user) => {
```
bobheadxi: did we decide on promises vs async-await?

web/components/login/SignedIn.js
```diff
@@ -0,0 +1,37 @@
...
+      }
+    });
+  }
+
+  render() {
+    let message = '';
+
+    if (this.state.user) {
+      // User is signed in.
+      message = `Signed in as ${this.state.user.displayName}`;
```
bobheadxi: 🔥

web/package.json
```diff
@@ -31,6 +31,8 @@
     "prop-types": "15.6.1",
     "react": "16.4.0",
     "react-dom": "16.4.0",
+    "react-firebaseui": "^3.0.4",
```
bobheadxi: Using `^` means that every time we install dependencies, the manager will grab the newest version - while semver ideally enforces that no breaking changes will occur, it's probably best to lock it down by removing `^` and only upgrading explicitly. This helps avoid `lock`file diffs polluting commit history.

bobheadxi:
Also reviewed the changes in the other PR because i didnt get a chance to - left some things that you might want to take note of: https://github.com/nwhacks/nwhacks2019/pull/3#pullrequestreview-131886969

web/components/login/SignedIn.js
mingyokim: nope

web/package.json
mingyokim: do we want to do that for every package in this file?

PR #5, opened 2018-06-27
Redux authentication [WIP]

mingyokim:


mingyokim:
Going with #8 instead

PR #6, opened 2018-06-27
Redux persist [WIP]

mingyokim:


mingyokim:
Replaced with #9 - implementing auth system without firebaseui

PR #9, opened 2018-07-01
Redux persist

mingyokim:
Setup `redux-persist` - I believe it re-loads the store faster when refreshing the page/navigating via url. (Don't have to reload the store by making requests to firebase - the store is persisted locally)


web/package.json
```diff
@@ -32,8 +32,10 @@
     "react-dom": "16.4.0",
     "react-redux": "5.0.7",
     "react-redux-firebase": "2.1.6",
+    "react-router": "^4.3.1",
```
bobheadxi: lock? 🔒 🐱

bobheadxi:
sweet

PR #10, opened 2018-07-01
Init admin panel

mingyokim:
Just a simple admin page where user access is restricted based on whether they are admin or not. 

Redirects to 404 if user is not logged in or not admin.

### DIFF
--- a/web/components/admin/index.js
+++ b/web/components/admin/index.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Link, Redirect } from 'react-router-dom';
+import { firebaseConnect } from 'react-redux-firebase';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+const AdminPanel = ({ auth }) => {
+  if (auth.isLoading) {
+    return (<span>Loading...</span>);
+  } else if (auth.isEmpty) {
+    return (
+      <Redirect to="/404" />
+    );
+  }
+
+  return (
+    <div>
+      <p>{ `Admin signed in as ${auth.displayName}` }</p>
+      <br />
+      <Link to="/logout">logout</Link>
+    </div>
+  );
+};
+
+AdminPanel.propTypes = {
+  auth: PropTypes.object,
+};
+
+export default compose(
+  firebaseConnect(),
+  connect(({ firebase: { auth } }) => { return { auth }; })
+)(AdminPanel);
--- a/web/components/admin/index.js
+++ b/web/components/admin/index.js
@@ -1,17 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Link, Redirect } from 'react-router-dom';
-import { firebaseConnect } from 'react-redux-firebase';
+import { firebaseConnect, firestoreConnect, isLoaded, isEmpty } from 'react-redux-firebase';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 
-const AdminPanel = ({ auth }) => {
-  if (auth.isLoading) {
+const AdminPanel = ({ firestore, auth }) => {
+  if (!isLoaded(auth)) {
     return (<span>Loading...</span>);
-  } else if (auth.isEmpty) {
-    return (
-      <Redirect to="/404" />
-    );
+  } else if (isEmpty(auth)) {
+    return (<Redirect to="/404" />);
+  }
+
+  const { admins } = firestore.data;
+
+  if (!isLoaded(admins)) {
+    return (<span>Loading...</span>);
+  } else if (firestore.errors.byQuery.admins) {
+    return (<Redirect to="/404" />);
   }
 
   return (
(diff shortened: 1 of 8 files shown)


web/components/admin/index.js
```diff
@@ -0,0 +1,46 @@
...
+import { Link, Redirect } from 'react-router-dom';
+import { firebaseConnect, firestoreConnect, isLoaded, isEmpty } from 'react-redux-firebase';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+
+const AdminPanel = ({ firestore, auth }) => {
+  if (!isLoaded(auth)) {
+    return (<span>Loading...</span>);
+  } else if (isEmpty(auth)) {
+    return (<Redirect to="/404" />);
```
bobheadxi: this should go to something like 403 right? (missing credentials)

bobheadxi: I also don't think we need brackets `()` around single-div jsx elements, starts to get really cluttered

web/components/admin/index.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const { admins } = firestore.data;
+
+  if (!isLoaded(admins)) {
+    return (<span>Loading...</span>);
```
bobheadxi: why can't this logic be in the same block as `isLoaded(auth)`?

web/components/admin/index.js
```diff
@@ -0,0 +1,46 @@
...
+  } else if (firestore.errors.byQuery.admins) {
+    return (<Redirect to="/404" />);
```
bobheadxi: this should be 401 (credentials refused)

web/components/login/Login.js
```diff
@@ -26,11 +26,15 @@ class Login extends React.Component {
...
+      const credential = {
         email: this.state.email,
         password: this.state.password,
-      }).then(() => {
+      };
+
+      this.props.firebase.login(credential).then(() => {
         this.setState({ loggedIn: true });
+      }).catch((error) => {
+        console.log(error);
```
bobheadxi: might want to do something like `this.setState({ errored: true})` for error message display instead of failing silently, though I'm not sure if we have plans for more global error handling

web/components/admin/index.js
mingyokim: I don't think we should let non-admin users know that an `admin` page exists at all so I think 404 will be better. What do you think?

bobheadxi: I don't think we need to be that secretive 😋 clarity is always nice, and if a user does stumble on the link, it's a bit more professional to say they're forbidden to access it than to say the page isn't found tbh. More importantly, could also help with tech support later on to differentiate between a bugged link, a missing account, or broken permissions, or whatever

> user: "can't get into the admin page"
> tech support: "what error are you getting?
> user: "404"
> tech support: "... 😨 "
> user: "... 😨 "

web/components/login/Login.js
mingyokim: Yeah i definitely agree! I created a separate issue for that here (https://www.notion.so/nwplus/Handle-Login-Error-6e3d864cca474b8bab4d21f936699fe6), didn't want to make this into a huge PR

web/components/admin/index.js
mingyokim: I would agree with you if users had different permissions levels and some users have access to admin page but some don't. But in our case, none of our users have access to the admin page. So why let them know about the admin page when none of them can access it anyway? 

Also, the admin page is not part of our "product/service" - it's more like an internal tool for our team. So it makes sense to me that users shouldn't know about the admin page.

@chamkank any thoughts?

mingyokim: Oh I’m not actually sending an HTTP code because we don’t have a server... should I be doing that? I’m just rendering a message basically.

bobheadxi: I don't think cham is saying we should be sending an HTTP code, just that the codes we display should match HTTP code standards if we're doing that.

> since our code is open source anyway, I don't see much value in not being specific with our codes.

👍

mingyokim: Hm still not convinced about the value in being specific about the error. 

Yes our code is open source, but our code is entirely front end - all our security implementations live on Firebase anyway. 

I think the question here is whether we want to help the logistics team when they somehow login incorrectly vs we make our app seem more robust by completely hiding features that users don’t need. 

We can make up for the lack of error description by training a team of 4 how to login and go on the admin page, if needed. But we can’t make up for the other way around. 

Also, if we somehow decide we need more error descriptions when landing on the admin page, we can always change the code to 403 and show error descriptions.

chamkank: My main issue with what you're proposing is that 404 is being used improperly here (not to avoid a security concern). If we're going to display HTTP codes, we should follow the standard.

> ... vs. we make our app seem more robust by completely hiding features that users don’t need

A workaround I can think of that takes both sides into account is having a generic error page (`Oops, something went wrong.`) where we can log the actual error using Sentry but not specify it on the UI. Thoughts?

bobheadxi: Yep - either use the HTTP codes properly or not at all, IMO.

mingyokim: Yeah, I'm not arguing that we should return 404 code. I just want to render the same page as what users would see if they enter an invalid/inexistent url. 

That being said, I'm totally for rendering a generic error here like "The page you're looking for doesn't exist" given that that's also what's rendered when users go into an invalid/inexistent url.

mingyokim: @bobheadxi @chamkank thoughts?

chamkank: +1 for generic error page without an HTTP error code specified in the UI

bobheadxi: generic error page sounds gud

bobheadxi:
good stuff 🔥

PR #11, opened 2018-07-05
Web configuration improvements

bobheadxi:
- removed package-lock
- updated dependencies
- `pages` are now `containers`
- new lint rules (the commit says `Adhere to new lint rules (airbnb 5.0.0)` but that's not really accurate oops), see 8f9311e - these are a lot more restrictive, but imo are quite nice. I recommend sticking with the new defaults
- cleaned up webpack configurations, adding some new loaders and hotloading


web/webpack.config.js
```diff
@@ -28,29 +31,27 @@ const config = {
...
         use: ['style-loader', 'css-loader', 'sass-loader'],
       },
+      {
+        // In case we use external libraries that have old-school CSS
+        test: /\.css$/,
+        use: ['style-loader', 'css-loader'],
+      },
+      {
+        // Package any static assets
+        test: /\.png($|\?)|\.jgp($|\?)|\.ico($|\?)|\.woff($|\?)|\.woff2($|\?)|\.ttf($|\?)|\.eot($|\?)|\.svg($|\?)/,
```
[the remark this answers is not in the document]
bobheadxi: 😨 that's pretty major

web/components/login/Login.js
```diff
@@ -9,11 +9,15 @@ class Login extends React.Component {
   constructor(props) {
     super(props);
 
+    // Deconstruct props and state
+    const { isLoaded, isEmpty, firebase, auth } = this.props;
+    const { email, password } = this.state;
```
mingyokim: This line needs to come after defining `this.state` in the line below

web/components/login/Login.js
```diff
@@ -9,11 +9,15 @@ class Login extends React.Component {
...
+
     this.state = {
       email: '',
       password: '',
-      loading: !this.props.auth.isLoaded,
-      loggedIn: !this.props.auth.isEmpty,
+      loading: !isLoaded,
```
mingyokim: We specifically want to check if `auth` is loaded; not sure if checking if the `props` is `isLoaded` encompasses whether `auth` is loaded. 

Regardless I think it would be best to be explicit here? by calling `isLoaded(auth)` and import `isLoaded` from `react-redux-firebase`

web/components/login/Login.js
```diff
@@ -9,11 +9,15 @@ class Login extends React.Component {
...
+      loggedIn: !isEmpty,
```
mingyokim: Same here, we probably want to specifically check if `auth` is loaded, not `props`

web/components/login/Login.js
bobheadxi: I want to avoid making any functional changes in this PR - here, `!isLoaded` does the same thing as what was there before (`!this.props.auth.isLoaded`)

web/components/login/Login.js
bobheadxi: nice catch, ill fix this

bobheadxi: see my other comment, this is just adapting the previous code to the new lint rules, I want to avoid making any non-cosmetic changes

web/components/login/Login.js
mingyokim: Sorry, i meant that `isLoaded` is not the same as what it was before. In the change above, you deconstructed `isLoaded` from `this.props` - meaning `isLoaded = this.props.isLoaded`. But we want it to be `isLoaded = this.props.auth.isLoaded`. 

So we should define `isLoaded` as `const { isLoaded } = auth` (or using some other ES6 syntax - i'm not too familiar with them) after line 11

bobheadxi: OH

bobheadxi: mb see 21b0c48

web/components/login/Login.js
bobheadxi: 21b0c48

bobheadxi:
Should probably look at 5ea3fe76b12baffe02397c2e04f5386416f6fbc9 as well for any code going forward - with destructuring, you can't assign functions that depend on state changes in the constructor, you'll have to load them in their respective functions each time. Instead of something like this:

```js
  constructor() {
    const { user } = this.state;
    this.signedInView = () => (
      <div>
        <p>Signed in as {user}</p>
        <Link to="/">Go to main</Link>
      </div>
    );
  }
```

You must write:

```js
  constructor() {
    this.signedInView = this.signedInView.bind(this);
  }

  signedInView() {
    const { user } = this.state;
    return (
      <div>
        <p>Signed in as {user}</p>
        <Link to="/">Go to main</Link>
      </div>
    );
  }
```

imo this is a lot cleaner than crowding the constructor with assignments.

See 5ea3fe7 and the associated commit description

mingyokim:
^ Does that actually increase performance or is it purely cosmetic (doesn't crowd constructor)?

bobheadxi:
As far as I know, mostly cosmetic and to avoid potential gotchas with how things load and update

mingyokim:
You need to move `NotFound.js` from `pages` to `containers`

bobheadxi:
goddammit travis

web/components/login/Login.js
```diff
@@ -8,53 +8,62 @@ import { connect } from 'react-redux';
...
-    };
+    this.onEmailChange = event => this.setState({ email: event.target.value });
+    this.onPasswordChange = event => this.setState({ password: event.target.value });
 
-    this.onPasswordChange = (event) => {
-      this.setState({ password: event.target.value });
-    };
+    this.login = this.login.bind(this);
+    this.loginView = this.loginView.bind(this);
+    this.signedInView = this.signedInView.bind(this);
```
d4l3k: What's the reason behind using `.bind(this)` here? Seems like signedInView is only used inside this class and thus you don't have any dynamic scope issues with `this`.

`login` needs to be bound since it's called from external functions, but not the internal view methods. `bind` is hideously slow and should be avoided whenever possible (90% slower).

https://jsperf.com/function-calls-direct-vs-apply-vs-call-vs-bind/14

bobheadxi: wow TIL I'll have a closer look at this, thanks!

bobheadxi: Where do you see `login()` being called externally? I think none of these functions need a `this` binding, since they're only called within this component, but maybe im misunderstanding

d4l3k: Might also consider using https://github.com/andreypopp/autobind-decorator instead. Makes the code more readable.

Could also try https://babeljs.io/docs/en/babel-plugin-transform-class-properties/ which lets you do `login = () => {}` at the class level.

bobheadxi: put up #14 to explore the babel plugin

PR #12, opened 2018-07-08
hide password input

mingyokim:
This PR hides the password input when logging in.


web/components/login/Login.js
```diff
@@ -54,6 +54,7 @@ class Login extends React.Component {
         <input
           placeholder="Enter your password"
           className="form-control"
+          type="password"
```
bobheadxi: amazing

PR #14, opened 2018-07-08
Use babel-plugin-transform-class-properties instead of bind

bobheadxi:

See https://github.com/nwhacks/nwhacks2019/pull/11#discussion_r200822229

This PR replaces `bind` calls with some `babel-plugin-transform-class-properties` magic

## :thought_balloon: Notes

This uses the `Stage 0` preset for babel-loader, which seems pretty broad but oh well

```
make web
```


web/components/login/Login.js
```diff
@@ -75,7 +71,7 @@ class Login extends React.Component {
     );
   }
 
-  signedInView() {
+  signedInView = () => {
```
d4l3k: The view functions don't need to be arrow functions right?

bobheadxi: well this one access `this.props` 🤷‍♂️

d4l3k: Yes? this.props works just fine since signedInView is called via this.signedInView().

() => {} explicity binds this to the function. This is mostly a style issue really

d4l3k: There's some performance penalties associated with using foo = () => {} over foo(){} since the first doesn't end up in the prototype and is an object property instead.

https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

Seems like @autobind is the most performant way to do this overall. Though, the performance difference is probably negligible anyways.

web/components/login/Login.js
```diff
@@ -19,10 +19,6 @@ class Login extends React.Component {
 
     this.onEmailChange = event => this.setState({ email: event.target.value });
     this.onPasswordChange = event => this.setState({ password: event.target.value });
```
d4l3k: Might consider moving this out of the constructor as well

web/components/login/Login.js
bobheadxi: oops yeah you're right about the view functions, I've fixed that 👍 thanks!

> Though, the performance difference is probably negligible anyways.

💭 

> We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil.

d4l3k:
If you're curious about what happens with all the different ways of doing function defs w/ binds here they are:

Source
```
import autobind from 'autobind-decorator'

class Component {
  constructor(value) {
    this.value = value

    this.constructorBind = () => {
      return this.value
    }
  }

  normal () {
    return this.value
  }

  arrow = () => {
    return this.value
  }

  @autobind
  bind() {
    return this.value
  }
}
```

Babel output
```
var _class;

function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key,
{ value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = v
alue; } return obj; }

function _applyDecoratedDescriptor(target, property, decorators, descriptor, context) { var de
sc = {}; Object['ke' + 'ys'](descriptor).forEach(function (key) { desc[key] = descriptor[key];
 }); desc.enumerable = !!desc.enumerable; desc.configurable = !!desc.configurable; if ('value'
 in desc || desc.initializer) { desc.writable = true; } desc = decorators.slice().reverse().re
duce(function (desc, decorator) { return decorator(target, property, desc) || desc; }, desc);
if (context && desc.initializer !== void 0) { desc.value = desc.initializer ? desc.initializer
.call(context) : void 0; desc.initializer = undefined; } if (desc.initializer === void 0) { Ob
ject['define' + 'Property'](target, property, desc); desc = null; } return desc; }

import autobind from 'autobind-decorator';
let Component = (_class = class Component {
  constructor(value) {
    _defineProperty(this, "arrow", () => {
      return this.value;
    });

    this.value = value;

    this.constructorBind = () => {
      return this.value;
    };
  }

  normal() {
    return this.value;
  }

  bind() {
    return this.value;
  }

}, (_applyDecoratedDescriptor(_class.prototype, "bind", [autobind], Object.getOwnPropertyDescr
iptor(_class.prototype, "bind"), _class.prototype)), _class);
```

bobheadxi:
ahhh yikes maybe autobind was the way to go, oh well

Thanks @d4l3k , this is all very insightful! will keep this in mind if we find the current method too slow

d4l3k:
The performance really doesn't matter in this case but I was just curious
what actually happens in the transformation

bobheadxi:
is the arrow function looks to be doing more or less the exact same thing as the constructor bind?

```js
    _defineProperty(this, "arrow", () => {
      return this.value;
    });

    this.constructorBind = () => {
      return this.value;
    };
```

d4l3k:
Yeah, pretty much. I was more looking at @autobind vs arrow implementation.
Autobind/decorators modify the prototype instead of the object properties
from what I can tell

PR #18, opened 2018-07-09
init secondary button

mingyokim:

This PR makes a `front_end_component` page (we will probably delete this later - this is just for visualizing the front end component without actually incorporating it into an existing component), makes a secondary button component that takes in `text` and `onClick` action. It also adds the `Apercu Pro` font. 

## :thought_balloon: Notes

Design: https://www.figma.com/file/KweWMEzkhgCpbpu5W9MZGE4L/nwHacks-2019-Website?node-id=66%3A435

Go to `/front_end_components` and open inspector - whenever you click on button, there should be a console log.


bobheadxi:
I think the `/front_end_components` page is a great idea btw, chuck all UI elements there for the time being

PR #19, opened 2018-07-10
Style navbar

mingyokim:

This one styles navbar according to design!!

image

## :thought_balloon: Notes

Design: https://www.figma.com/file/KweWMEzkhgCpbpu5W9MZGE4L/nwHacks-2019-Website?node-id=53%3A2

DO NOT MERGE until #18 is merged!

Should behave like the wireframes in design


mingyokim:
@bobheadxi thanks, i'm a css noob so any feedback is appreciated!

also let me know if you think it'd be worth to make a link component instead of just styling `a` directly (i think there are a lot of those in the design)

PR #20, opened 2018-07-10
Primary button, stylesheet restructure

bobheadxi:

- Primary button based on figma design
- Moved `frontendcomponents` into `demo`
- Reorganized button styles
- Reorganized stylesheet imports

primarybutton

## :thought_balloon: Notes

For stylesheets, there should only be one import for general styles in `index.js`:

```js
import React from 'react';
import ReactDOM from 'react-dom';

import App from './components/app/App';
import './styles/index.sass';

ReactDOM.render(<App />, document.getElementById('app'));
```

All other styles should be imported via `styles/index.sass`:

```sass
/* core */
@import ./colors
@import ./typography

/* components */
@import ../components/index

/* containers */
@import ../containers/index
```

etc etc.

```
make web
```

then go to http://localhost:8080/ui_demo


web/components/input/buttons/PrimaryButton/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PrimaryButton = ({ text, onClick, disabled }) => (
+  <button
+    onClick={onClick}
+    type="button"
+    className={`primary ${disabled ? 'disabled' : ''}`}>
```
mingyokim: Can we disable buttons by doing `<button disabled>Button</button>` instead?
https://stackoverflow.com/questions/14750078/style-disabled-button-with-css

bobheadxi: That's much cleaner thanks, addressed in 7b2f778

web/components/demo/index.js
```diff
@@ -0,0 +1,25 @@
...
+    <p>Buttons</p>
+    <br />
+    <div>
+      <PrimaryButton text="Primary" onClick={ButtonCallback} />
+      &nbsp;
+      <SecondaryButton text="Secondary" onClick={ButtonCallback} />
+    </div>
+    <br />
+    <div>
+      <PrimaryButton text="Primary" onClick={ButtonCallback} disabled />
```
mingyokim: I would like to see some specs that ensures `onClick` callback is not called when it's `disabled`, and that it's called when it's not `disabled`

bobheadxi:
@mingyokim 

> refactor them into one component, and maybe pass in an additional prop like “type” that specifies whether the button is primary/secondary

I think that's a pretty bad idea, and having them be distinct classes is useful and pretty in line with how I've seen React components being used, otherwise we might as well forgo the React component and just use a bare `<button ... />` div when we need buttons.

web/components/input/buttons/PrimaryButton/PrimaryButton.sass
```diff
@@ -0,0 +1,7 @@
+button.primary
+  @include gradient-hover-animation($gradient-primary, $gradient-primary-highlight)
```
mingyokim: 🎉

web/components/input/buttons/SecondaryButton/SecondaryButton.sass
```diff
@@ -0,0 +1,13 @@
+button.secondary
+  background: $gradient-secondary
```
mingyokim: We should make the styling for `SecondaryButton` the same as `PrimaryButton` either in this PR or create a follow-up issue. 

OR we could refactor `PrimaryButton` and `SecondaryButton` into one component as I've already mentioned

web/components/input/buttons/index.js
```diff
@@ -0,0 +1,7 @@
+import PrimaryButton from './PrimaryButton';
```
mingyokim: Question about the location of the button component: is button an input?

web/styles/animations.sass
```diff
@@ -0,0 +1,30 @@
+$transition-time: .2s
+
+// This is a hack to allow transitions of linear-gradient backgrounds.
+// https://medium.com/@dave_lunny/animating-css-gradients-using-only-css-d2fd7671e759
+// Use this by @include'ing it in your class with required arguments.
+// Arguments:
+//   $main        linear-gradient to use as normal background
+//   $hover       linear-gradient to transition to upon hover
+//   $transition  [optional] transition time, defaults to $transition-time
+@mixin gradient-hover-animation($main, $hover, $transition: $transition-time )
```
mingyokim: ...OK i trust you

web/styles/colors.sass
```diff
@@ -0,0 +1,20 @@
+/* colors */
+$white: #FFFFFF
```
mingyokim: yaaas we need this - we should also consider having a set variables for spacing as well.

@sherryyx is there currently a spacing schema?

web/styles/index.sass
```diff
@@ -1,5 +1,13 @@
+/* core */
 @import ./colors
 @import ./typography
+@import ./animations
+
+/* components */
+@import ../components/index
```
mingyokim: 💪i like how the order of import is really clear here

web/components/input/buttons/index.js
bobheadxi: Absolutely! The [standard way to declare buttons used to be `<input type="button" />`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/button)

mingyokim:
@bobheadxi Sure, we can just leave it as two separate components until we start seeing a lot of repetitions - which as you've mentioned is probably unlikely.

web/components/input/buttons/SecondaryButton/SecondaryButton.sass
mingyokim: Ok we've decided not to refactor it, but i'd still like to see consistent `sass` between the two components? Like I'm not seeing the `@include gradient-hover-animation($gradient-primary, $gradient-primary-highlight)` piece here

bobheadxi: What do you mean about the styling? I might agree on making them the same with only colour variations, but the special CSS required to make `linear-gradient` transition properly sorta makes that more effort than its worth IMO 😞

web/styles/colors.sass
mingyokim: (^ not part of this PR) - something to consider going forward

web/components/input/buttons/SecondaryButton/SecondaryButton.sass
bobheadxi: > Like I'm not seeing the @include gradient-hover-animation($gradient-primary, $gradient-primary-highlight) piece here

The secondary button transitions to a flat color `$secondary-highlight`, which works with `-webkit-transition`, but the primary button's `gradient`-to-`gradient` transition doesn't rip, hence the need for the mixin on one and not the other.

web/components/demo/index.js
bobheadxi: I'll look into adding a few simple tests for this PR, though I'll just point out that the disabled-ness is pretty enforced in the CSS:

```sass
  &:disabled
    pointer-events: none
```

web/components/input/buttons/SecondaryButton/SecondaryButton.sass
bobheadxi: A lot of the shared styles are already in `index.sass` of `buttons`, which worked pretty well since they pretty much all applied to the progress group buttons as well (see the CSS in #21)

bobheadxi: Thought: we should be careful about making the styling too tightly coupled between the components, because it makes us really inflexible when changes are requested - it's easier to copy styles across multiple components than it is to disentangle a property from tightly coupled styles.

web/components/demo/index.js
mingyokim: Yeah I get that's enforced in the CSS but we can't predict how the CSS file would change later on - tests help us ensure that our components are still working as expected no matter how much our code changes in the future.

web/components/input/buttons/SecondaryButton/SecondaryButton.sass
mingyokim: Okay nevermind, i see what you mean

and yeah i'm a noob to css so i don't fully understand the pros/cons of the two approaches but i think if we have a fully fleshed out style guide then it's possible to achieve styles that are tightly coupled between the components

i guess let's bring it up in our channel/next meeting?

bobheadxi: 🚀

web/components/demo/index.js
bobheadxi: See 941495c

web/components/input/buttons/PrimaryButton/PrimaryButton.test.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    test('should have text', () => {
+      expect(wrapper.text()).toEqual('hello cham');
+    });
+
+    test('should not be clickable', () => {
+      expect(wrapper.find('button').props().disabled).toBeTruthy();
+
+      // Enzyme click simulations do not respect 'disabled' attribute.
```
mingyokim: 😞

web/components/input/buttons/PrimaryButton/PrimaryButton.test.js
```diff
@@ -0,0 +1,59 @@
...
+      });
+    });
+
+    test('should have text', () => {
+      expect(wrapper.text()).toEqual('hello cham');
+    });
+
+    test('should be clickable', () => {
+      wrapper.find('button').simulate('click');
+      expect(clicked).toBeTruthy();
```
mingyokim: 🎉

PR #21, opened 2018-07-11
Progress button group

bobheadxi:
Lunch break 🤷‍♂️ 

progressgroup

## :thought_balloon: Notes

~Not sure what best to do for inputs to generate the component~

~Current usage:~

```js
      <ProgressGroup
        activeIndex={1}
        steps={[
          { text: '1', onClick: ButtonCallback },
          { text: '2', onClick: ButtonCallback },
          { text: '3', onClick: ButtonCallback, disabled: true },
          { text: '4', onClick: ButtonCallback, disabled: true },
        ]} />
```

~PR currently pointed at #20 for a more accurate diff.~

```
make web

# go to http://localhost:8080/ui_demo
```

### DIFF
--- a/web/components/demo/index.js
+++ b/web/components/demo/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { SecondaryButton, PrimaryButton } from '../input/buttons';
+import { SecondaryButton, PrimaryButton, ProgressGroup } from '../input/buttons';
 
 const ButtonCallback = e => console.log(`${e.currentTarget.textContent} button clicked!`);
 
@@ -18,6 +18,17 @@ const FrontEndComponents = () => (
       <PrimaryButton text="Primary" onClick={ButtonCallback} disabled />
       &nbsp;
       <SecondaryButton text="Secondary" onClick={ButtonCallback} disabled />
+    </div>
+    <br />
+    <div>
+      <ProgressGroup
+        activeIndex={1}
+        steps={[
+          { text: '1', onClick: ButtonCallback },
+          { text: '2', onClick: ButtonCallback },
+          { text: '3', onClick: ButtonCallback, disabled: true },
+          { text: '4', onClick: ButtonCallback, disabled: true },
+        ]} />
     </div>
   </div>
 );
--- a/web/components/demo/index.js
+++ b/web/components/demo/index.js
@@ -4,33 +4,45 @@ import { SecondaryButton, PrimaryButton, ProgressGroup } from '../input/buttons'
 
 const ButtonCallback = e => console.log(`${e.currentTarget.textContent} button clicked!`);
 
-const FrontEndComponents = () => (
-  <div>
-    <p>Buttons</p>
-    <br />
-    <div>
-      <PrimaryButton text="Primary" onClick={ButtonCallback} />
-      &nbsp;
-      <SecondaryButton text="Secondary" onClick={ButtonCallback} />
-    </div>
-    <br />
-    <div>
-      <PrimaryButton text="Primary" onClick={ButtonCallback} disabled />
-      &nbsp;
-      <SecondaryButton text="Secondary" onClick={ButtonCallback} disabled />
-    </div>
-    <br />
-    <div>
-      <ProgressGroup
-        activeIndex={1}
-        steps={[
-          { text: '1', onClick: ButtonCallback },
-          { text: '2', onClick: ButtonCallback },
-          { text: '3', onClick: ButtonCallback, disabled: true },
-          { text: '4', onClick: ButtonCallback, disabled: true },
-        ]} />
-    </div>
-  </div>
-);
+class FrontEndComponents extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      active: 3,
+    };
+  }
+
+  switchProgress = e => this.setState({ active: parseInt(e.currentTarget.textContent, 10) - 1 })
+
+  render() {
+    const { active } = this.state;
+    return (
+      <div>
+        <p>Buttons</p>
+        <br />
+        <div>
+          <PrimaryButton text="Primary" onClick={ButtonCallback} />
+          &nbsp;
+          <SecondaryButton text="Secondary" onClick={ButtonCallback} />
+        </div>
+        <br />
+        <div>
+          <PrimaryButton text="Primary" onClick={ButtonCallback} disabled />
+          &nbsp;
+          <SecondaryButton text="Secondary" onClick={ButtonCallback} disabled />
+        </div>
+        <br />
+        <div>
+          <ProgressGroup
+            count={10}
+            onClick={this.switchProgress}
+            activeIndex={active}
+            lastValidIndex={7}
+          />
+        </div>
+      </div>
+    );
+  }
+}
 
 export default FrontEndComponents;
--- a/web/components/demo/index.js
+++ b/web/components/demo/index.js
@@ -12,7 +12,7 @@ class FrontEndComponents extends React.Component {
     };
   }
 
-  switchProgress = e => this.setState({ active: parseInt(e.currentTarget.textContent, 10) - 1 })
+  switchProgress = index => this.setState({ active: index })
 
   render() {
     const { active } = this.state;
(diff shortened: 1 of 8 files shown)


web/components/demo/index.js
```diff
@@ -19,6 +19,17 @@ const FrontEndComponents = () => (
...
       <SecondaryButton text="Secondary" onClick={ButtonCallback} disabled />
     </div>
+    <br />
+    <div>
+      <ProgressGroup
+        activeIndex={1}
+        steps={[
+          { text: '1', onClick: ButtonCallback },
+          { text: '2', onClick: ButtonCallback },
+          { text: '3', onClick: ButtonCallback, disabled: true },
```
mingyokim: Can we move the disable logic to the progress group component? Ideally we would only pass in activeIndex and number of buttons. And just disable all buttons with index larger than activeIndex

bobheadxi: I was thinking about this too, but `activeIndex` could be less than the most activate-able index. For example, you could get to step 3, then go to step 2, but you want step 3 to still be clickable.

What do you think about:

```js
ProgressGroup.propTypes = {
  activeIndex: PropTypes.number, // "clicked"
  lastActivatableIndex: PropTypes.number, // all values after this will be "disabled"
  steps: PropTypes.arrayOf(PropTypes.shape({
    text: PropTypes.string.isRequired,
    onClick: PropTypes.func,
  })),
};
```

mingyokim: I don’t think steps is necessary - you can automatically generate n number of buttons labeled from 1 to n, given that you pass in n in the props. And we’ll need a callback function that is called whenever the page is changed so that whoever is listening to that callback can know what page it’s in.

And whether we can click a page that’s after the current page is something we should confirm with design

bobheadxi: I'll go with my second idea then, and just allow one callback that all the button clicks call which provides the current page index.

bobheadxi:
progressgroup-2

web/components/input/buttons/ProgressGroup/index.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const ProgressGroup = ({ count, onClick, activeIndex, lastValidIndex }) => {
+  const buttons = [];
+  for (let i = 0; i < count; i += 1) {
```
mingyokim: Why are we 0-indexing here? Wouldn't it make more sense to start from one since page numbers also start with 1? I think it would be easier that way

web/components/input/buttons/ProgressGroup/index.js
```diff
@@ -0,0 +1,32 @@
...
+    const isActive = (i === activeIndex);
+    buttons.push(
+      <button
+        key={i}
+        onClick={onClick}
```
mingyokim: This API is definitely an improvement, but I'm still not digging https://github.com/nwhacks/nwhacks2019/pull/21/files#diff-35b1e54975e81f41e3665bf4708f4329R15 - where you have to get the page number by reading the text value of the element (`parseInt(e.currentTarget.textContent, 10) - 1`). We want to get the page number/index as the argument, nothing more. 

We could easily account for that by declaring another callback function like so:
`const func = () => onClick(i)`
and passing in this function as the onClick function:
`onClick={func}`

bobheadxi: > We want to get the page number/index as the argument, nothing more.

ah, was lazy. Good idea

web/components/input/buttons/ProgressGroup/index.js
bobheadxi: 1-indexing is a sin!

bobheadxi: tbh though 0-indexing is better here - the text in the button is purely cosmetic, and when you are writing up the logic it makes more sense to start at 0 to match array loops, etc.

web/components/input/buttons/ProgressGroup/index.js
bobheadxi: 886cfcd

web/components/input/buttons/ProgressGroup/index.js
mingyokim: I still don't really understand why 0-indexing is better but can we at least change the API so that the user doesn't have to know anything about "index"? So changing `activeIndex` to `activePage` and `lastValidIndex` to `lastValidPage`? Don't really see why the user has to know anything about how the buttons are stored.

bobheadxi: The only users are us, not the users of the site, and as far as writing code goes 0-indexing is correct and anything else is honestly confusing - all lists start at 0, why should this list start at 1?

http://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html

bobheadxi: Users don’t have to know how it’s stored but we do

mingyokim: The users are not only us - the users include devs who do not know any implementation detail of this component (e.g. new devs next year who may want to reuse our components, or those who weren’t involved in developing this component but need to use it to build another component). When writing base components it’s really important to think about reusability and good API to ultimately save time in the future. 

By forcing users to specify an index, we are revealing a detail of our implementation which is completely unnecessary. All they want to do is generate progress buttons from page 1 to n and specify the current page and last active page. 

And about 0-indexing i don’t really care but if we do go forward with changing the API to specify page instead of index, you’d literally have to do `i+1` whenever you reference `i`. So starting with 0 seems especially unnecessary compared to just starting with 1. Moreover we are not really “storing” the buttons in a list, we’re just rendering a list of buttons using a for loop. If the for loop started with 1, the first button would still be 0-indexed - `buttons[0]` - so wouldn’t be doing anything crazy here. Same old 0-indexed list, but generated with a for loop that starts at 1 instead of 0.

bobheadxi: >  When writing base components it’s really important to think about reusability and good API to ultimately save time in the future.

0-index will do exactly that, it is the standard all APIs are based on.

> By forcing users to specify an index, we are revealing a detail of our implementation which is completely unnecessary. All they want to do is generate progress buttons from page 1 to n and specify the current page and last active page.

Thats a really weird way of looking at it - I want to generate a set of n elements, and iterate through them.

> And about 0-indexing i don’t really care but if we do go forward with changing the API to specify page instead of index, you’d literally have to do i+1 whenever you reference i.

What? Why would you specify page by the name? Just reference the order. API is simple enough: https://github.com/nwhacks/nwhacks2019/pull/21/files#diff-35b1e54975e81f41e3665bf4708f4329R15

> Moreover we are not really “storing” the buttons in a list, we’re just rendering a list of buttons using a for loop.

This is a list! A list of pages. If I want the first of an list, I want the 0th of that list. I imagine most developers would use that standard. I've literally never seen any API of any sort start with a 1.

> but generated with a for loop that starts at 1 instead of 0.

You should read the article I linked - if we start at 1, why not 10, or 50? 0 is the 0 value, it is the beginning 🥇 

Also things like:
0 - 1 = -1, obviously invalid
1 - 1 = 0, .... is that invalid? but all code uses the 0-index, why is this invalid?

This is all really philosophical. In summary: 0-index is the standard, it makes sense, users expect it in APIs (I never expect a 1-indexed API)... there's literally no reason to index these by their name.

bobheadxi: > the users include devs who do not know any implementation detail of this component 

these devs will expect a 0-index, because that's just how things work lol. It's the 1-index that will stump anyone who has experience with working with code and APIs.

mingyokim:
@bobheadxi whoops sorry I should've approved (i was on my phone) - i'm okay with either approach since the functionality is the same as you've mentioned.

bobheadxi:
All good, thanks for the comments!

mingyokim:
@bobheadxi wait i forgot to do a sanity check before you merged it, and it doesn't seem like `lastActiveIndex` is working - on `ui_demo` page, I can still click the page 9 and 10? Can you confirm this bug?

bobheadxi:
Ah, I forgot to change `disabled` from a class to a button trait after merging from the other PR. I'll put up a patch soon (and write some tests 🤕 )

mingyokim:
Okay thanks! And I agree, would definitely love to see some tests on this component (y)

bobheadxi:
Tracking in #25

PR #24, opened 2018-07-17
Click state

bobheadxi:

button-press

## :thought_balloon: Notes

this was such a pain in the ass ughhhh and there are still some bugs (see gif...)

currently still investigating

make web -> /ui_demo


mingyokim:
I think worse case scenario we could just use pure CSS like this: https://codepen.io/finnhvman/pen/jLXKJw but without having the ripple start at the cursor

mingyokim:
@bobheadxi this works perfectly fine on my machine... i don't see the bug in the gif

mingyokim:
i also noticed this is almost the same as https://github.com/rwu823/react-ripples/blob/master/src/index.js - any reason why we're not using this package?

also another thing is the ripple isn't always consistent - it sometimes doesn't ripple when u click

bobheadxi:
The animation logic is certainly inspired, but I'm not a fan of how it's done there. My changes didnt work too well, so yeah current state of the PR is quite similar to the code in that example. I'll have another go this weekend.

PR #25, opened 2018-07-19
Fix disabled progress buttons, add tests

bobheadxi:

Fix disabled progress buttons, add tests - still WIP

## :thought_balloon: Notes

n/a

```
make test
```


bobheadxi:
Reminder on this @chamkank @mingyokim

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    test('all buttons should be clickable', () => {
+      for (let i = 0; i < count; i += 1) {
+        wrapper.childAt(i).simulate('click');
+        expect(clicked).toEqual(i);
+      }
+    });
+
+    test('active button should have class "active"', () => {
+      expect(wrapper.childAt(activeIndex).hasClass('active')).toBeTruthy();
```
mingyokim: We could also iterate through all the buttons to check that the non-active buttons don't have the class `active`

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
```diff
@@ -0,0 +1,84 @@
...
+    });
+  });
+
+  describe('when some steps are disabled', () => {
+    const count = 4;
+    const lastActiveIndex = 4;
```
mingyokim: I'm confused... why is `lastActiveIndex` 4? The max it can be should be 3 based on `count`

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
```diff
@@ -0,0 +1,84 @@
...
+      for (let i = 0; i < count; i += 1) {
+        expect(wrapper.childAt(i).text()).toEqual((i + 1).toString());
+      }
+    });
+
+    test('some buttons should not be clickable', () => {
+      for (let i = 0; i < count; i += 1) {
+        wrapper.childAt(i).simulate('click');
+        if (i <= lastActiveIndex) {
+          expect(clicked).toEqual(i);
```
mingyokim: we could also test that the component has `disabled` property, especially since "Enzyme click simulations do not respect 'disabled' attribute" as you've mentioned.

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
bobheadxi: I don't think it's very important - makes it easy to enable all buttons by setting lastActiveIndex as 999

mingyokim: oh yeah but i think this test is trying to test when some steps are disabled - so why is it enabling all buttons?

bobheadxi: oh wait, this isn't used at all:

```js
     wrapper = getWrapper({
        count,
        onClick: (i) => { clicked = i; },
        activeIndex: 1,
        lastActiveIndex: count / 2,
      });
```

ill remove it oops

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
bobheadxi: that's on the next line:

```js
        if (i <= lastActiveIndex) {
          expect(clicked).toEqual(i);
        } else {
          expect(wrapper.childAt(i).props().disabled).toBeTruthy();

          // Enzyme click simulations do not respect 'disabled' attribute.
          // See https://github.com/airbnb/enzyme/issues/386
          // expect(clicked).not.toEqual(i)
        }
```

mingyokim: oh sorry, i meant to test that the component does NOT have `disabled` property when it's less than `lastActiveIndex`

i.e. `expect(wrapper.childAt(i).props().disabled).toBeFalsy();`

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
```diff
@@ -0,0 +1,101 @@
...
+      for (let i = 0; i < count; i += 1) {
+        expect(wrapper.find('button').at(i).text()).toEqual((i + 1).toString());
+      }
+    });
+
+    test('some buttons should not be clickable', () => {
+      for (let i = 0; i < count; i += 1) {
+        wrapper.find('button').at(i).simulate('click');
+        if (i <= lastValidIndex) {
+          expect(clicked).toEqual(i);
```
mingyokim: can we test that the prop does not have disabled property when it's less than lastActiveIndex?

i.e. `expect(wrapper.childAt(i).props().disabled).toBeFalsy();`

I think it's especially important here since enzyme click simulations don't work with `disabled` attribute - `onClick` can be triggered even if the button is `disabled`, so testing that `onClick` is triggered by itself would not guarantee that the button is not `disabled`

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
bobheadxi: sure!

web/components/input/buttons/ProgressGroup/ProgressGroup.test.js
bobheadxi: true - see fa0dd52

bobheadxi:
Thanks for the feedback @mingyokim

PR #28, opened 2018-07-26
init hacker application route

mingyokim:

Introduces `application/hacker` route! It also calls Firestore to determine whether the user has already applied or not (if logged in).

## :thought_balloon: Notes

This PR is also the beginning of storing essential user information into redux on login, and destroying the store on logout. In this PR, we store the hacker application into the store as the user logs in, so that we have that hacker application ready when the login redirects to dashboard. We could also do the same for determining if the user is admin, and a lot of other handy stuff that we know we'll need for sure.

### User has already submitted a hacker application
* create a user on firebase authentication
* create a doc in `applications_hacker` with the user's `uid` as the `id`
* Login as that user
- [ ] dashboard says "you have one hacker application"
* go to `application/hacker`
- [ ] you are redirected back to `/dashboard`

### User has not already submitted a hacker application
* create a user on firebase authentication
* ensure there's no doc in`applications_hacker` whose `id` matches the user's `uid`
* Login as that user
- [ ] dashboard says "you didn't apply as a hacker"
* go to `application/hacker`
- [ ] you are not redirected to `/dashboard` and tells you to apply

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -19,7 +19,7 @@
       "\\.(css|less|scss|sass)$": "identity-obj-proxy"
     },
     "collectCoverage": true,
-    "collectCoverageFrom" : [
+    "collectCoverageFrom": [
       "**/*.js",
       "!**/node_modules/**",
       "!**/*.test.js",
@@ -27,7 +27,10 @@
       "!**/coverage/**",
       "!**/tests/**"
     ],
-    "coverageReporters": ["text", "json"]
+    "coverageReporters": [
+      "text",
+      "json"
+    ]
   },
   "dependencies": {
     "@firebase/app": "0.3.3",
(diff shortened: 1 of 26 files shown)


web/package.json
```diff
@@ -19,15 +19,18 @@
       "\\.(css|less|scss|sass)$": "identity-obj-proxy"
     },
     "collectCoverage": true,
-    "collectCoverageFrom" : [
+    "collectCoverageFrom": [
```
bobheadxi: heh

web/services/store/index.js
```diff
@@ -11,11 +11,9 @@ import localStorage from 'redux-persist/lib/storage';
 import hardSet from 'redux-persist/lib/stateReconciler/hardSet';
 
 import { firebaseConfig, rrfConfig } from '../../main.config';
+import Reducers from '../../reducers';
 
-const persistConfig = {
-  key: 'root',
-  storage: localStorage,
-};
+import { persistConfig } from './persist.config';
```
bobheadxi: any reason this is being moved out? it doesnt look like `persistConfig` is referenced elsewhere

mingyokim:
@chamkank good catch, will address in another pr!

PR #29, opened 2018-08-01
init/style Checkbox & CheckboxGroup

chamkank:

This PR adds the `Checkbox` and `CheckboxGroup` components. 

#### Usage:
New checkbox demo

You can use `Checkbox` on its own but it's cleaner when wrapped with `CheckboxGroup` because you can have a shared `name` attribute (as well as shared `defaultChecked`, `disabled`, and `onChange` attributes). You can override a shared prop by passing a specific value for that prop to the child component.
```
import React from 'react';
import { Checkbox, CheckboxGroup } from '../input/buttons/CheckboxGroup';

<CheckboxGroup name="test_name">
  <Checkbox label="Label 1" value="test_value" checked="checked"/>
  <Checkbox label="Label 2" value="test_value" disabled />
  <Checkbox label="Label 3" value="test_value" name="overriden_name"/>
</CheckboxGroup>
```

## :thought_balloon: Notes
<details>
  <summary>If you're curious as to what the pure CSS implementation looked like...</summary>

Old checkbox
 - Close but not close enough to the mocks, so I used the SVG images instead.

</details>

### TODO
- [ ] I still need to add styling for the disabled state
- [ ] Also need to add Enzyme tests

`make web` & go to `localhost:8080/ui_demo`


web/assets/checkbox/checked.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
bobheadxi: `xmlns` is XML namespacing, which iirc is used to do the svg magic - all svg's will have it

web/components/input/buttons/CheckboxGroup/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const InputContext = React.createContext();
+
+const Checkbox = ({ name, value, label, checked, disabled, onChange }) => (
+  <InputContext.Consumer>
```
mingyokim: This is pretty cool, never seen this pattern before.

Would you still be able to render a `Checkbox` without `CheckboxGroup`?

web/components/input/buttons/CheckboxGroup/index.js
```diff
@@ -0,0 +1,54 @@
...
+    {context => (
+      <label className="checkbox">
+        <input
+          type="checkbox"
+          name={name || context.name}
+          value={value}
```
mingyokim: Why do we need `value`? Couldn't `onChange` just supply the value in the callback?

web/components/input/buttons/CheckboxGroup/index.js
```diff
@@ -0,0 +1,54 @@
...
+    )}
+  </InputContext.Consumer>
+);
+
+Checkbox.propTypes = {
+  // name should be the same amongst checkboxes in a checkbox group
+  name: PropTypes.string,
+
+  // input value for checkbox
+  value: PropTypes.string.isRequired,
```
mingyokim: wouldn't this be a boolean?

web/components/input/buttons/CheckboxGroup/index.js
chamkank: IKR, it's the [new context API](https://reactjs.org/docs/context.html) and I like it a lot. :smiley: 

I intended for `Checkbox` to be used on its own but currently it's failing without a parent `CheckboxGroup`. I'll fix this tomorrow, thanks!

web/components/input/buttons/CheckboxGroup/index.js
chamkank: Not sure if I follow. Normally in regular forms you can have an input without an `onChange` function (the value is passed on form submit) and so I wanted to support that use case. I haven't played around with forms in React before, is there a reason to have the `onChange` without the `value`?

web/components/input/buttons/CheckboxGroup/index.js
chamkank: Which part are you referring to?

bobheadxi: checkbox value is either true or false, right? unless value is the label, but it looks like that's covered by `name || context.name`

web/components/input/buttons/CheckboxGroup/index.js
bobheadxi: I think this goes for any checkbox form - checkbox is just a toggle, `onChange` just needs to notify if value is `true` or `false`

chamkank: OH SHOOT, I've mixed up my logic, the shared `name` prop is only useful to radio buttons (where there is only a single answer), hence why `value` is there so you can distinguish between radio button options. 

But this is a checkbox... :skull: , I'll fix all of this tomorrow.

bobheadxi: That's fair actually, I think you're right

chamkank:
Bad news lol, `enzyme-adapter-react-16` doesn't support the new context API yet. Stumbled upon it while writing tests (see [this error](https://github.com/airbnb/enzyme/issues/1620)). Good news is that it's being worked on (see [this issue](https://github.com/airbnb/enzyme/issues/1553)). So no tests for this (and the radio button) until that's fixed.

web/.eslintrc.json
```diff
@@ -17,6 +17,7 @@
     "react/jsx-filename-extension": 0,
     "react/require-default-props": 0,
     "react/forbid-prop-types": 0,
+    "react/destructuring-assignment": 0,
```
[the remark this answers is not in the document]
chamkank: I fixed `CheckboxGroup` to use destructuring (`const { children, ...rest } = props;`) but that broke another rule which was `no-unused-vars` (since `...rest` wasn't used).

But now I realize that `...rest` is optional :stuck_out_tongue: , I'll make this change.

web/.eslintrc.json
```diff
@@ -47,6 +47,22 @@
...
       }
+    ], 
+    "react/prop-types": [
+      "enabled",
+      { 
+        "ignore": "ignore", 
+        "customValidators": "customValidator"
+      }
+    ],
+    "jsx-a11y/label-has-for": [
```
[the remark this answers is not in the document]
chamkank: I've removed the `prop-types` config (it allowed me to not have to write `propTypes` for `CheckboxGroup`). 

The `jsx-a11y/label-has-for` config was for dealing with the error related to not having a `for` attribute for the label. There are two ways I know of to associate labels with checkboxes (which makes the labels clickable):
 
```
<input id="some_id">
<label for="some_id">Some label<label>
```
```
<label>
  <input /> Some label
</label>
```
And I chose the second option. With the first one we have to either require that a unique `id` is passed in to each `Checkbox` (or we generate one for each Checkbox if it's not provided) which seemed unnecessary. The config I added allows us to use either approach - without the config it forces us to use both, which would look like this:

```
<label for="some_id">
  <input id="some_id" /> Some label
</label>
```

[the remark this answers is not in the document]
bobheadxi: > (it allowed me to not have to write propTypes for CheckboxGroup).

Why would you not want this?

PR #30, opened 2018-08-02
Web/build

bobheadxi:

Recommend reviewing this commit-by-commit.

Saw some stuff in the other PRs where images, assets were `src`'d via relative paths, which won't work unless we keep everything in `docs/` and keep the paths set up right... blah blah it's a bit of a hassle. So we let webpack handle it:

```js
import logo from '../../assets/logo.png';

const getLogo = () => (
  <Link to="/"><img alt="" src={logo} /></Link>
);
```

this is set in the webpack config:

```js
      {
        // Package any static assets
        test: /\.png($|\?)|\.jpg($|\?)|\.ico($|\?)|\.otf($|\?)|\.woff($|\?)|\.woff2($|\?)|\.ttf($|\?)|\.eot($|\?)|\.svg($|\?)/,
        loader: 'url-loader',
      },
```

while I was at it, I did some other adjustments and whatnot to the makefile, and added a script to emulate serving a static web app, just in case webpack-dev-server behaves differently (ie the logo example above)

also added a thing to make sure that react-router works properly with gh-pages based on a ticket I found, but we'll have to keep an eye on if it works. More details in corresponding commit

also bumped dependencies

```bash
make build
make serve
# go to localhost:8081
```


Makefile
```diff
@@ -1,43 +1,84 @@
...
 web:
-	(cd ./web ; yarn start)
+	@echo Starting webpack-dev-server...
+	@(cd ./web ; yarn start)
+
+# Runs a simple web serve to test the build static web app
+.PHONY: serve
+serve:
+	@echo Go to http://localhost:8081 to test the built web app
+	@(cd ./docs ; python ../serve.py)
```
chamkank: Can we use [http-server](https://www.npmjs.com/package/http-server) instead of the Python script? I know Python 2 & 3 come installed by default on Macs but that's not the case for Windows.  After running `yarn add http-server -D` in web, we can replace line 60 in the Makefile with `@(./web/node_modules/.bin/http-server ./docs -a localhost -p 8081 -c-1)`.

Makefile
```diff
@@ -1,43 +1,84 @@
 all: deps
 
+# List all commands
+.PHONY: ls
```
chamkank: Woohoo this will be useful!

Makefile
chamkank: That would be nice because requiring the user to have Python 2 installed is more of an inconvenience than installing an additional npm package in my mind. Also avoids having to create a new file just for serving the app. Is there a reason why we wouldn't do mock data filling in JavaScript? Imo it would be really nice to stick to one language unless we have a compelling reason not to.

bobheadxi: No reason, one language is a bit boring and I just find python more convenient, and python2 is here because as you mentioned its installed by default on osx. But javascript is fine too.

Fair point on the extra file (though extra npm dependency 😢 ) - i'll make the changes tonight to use the npm package instead

bobheadxi: 542a586

PR #38, opened 2018-08-09
init RadioButton RadioGroup components

chamkank:

This adds the `RadioButton` and `RadioGroup` components. #37 

## :thought_balloon: Notes

Pretty much identical to my implementation of the checkbox components but I'd rather not couple the two to use the same stylesheet.

`make web` & go to `localhost:8080/ui_demo`


web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -0,0 +1,54 @@
...
+    background-image: url('../assets/radio/unselected.svg')
+
+    &:after
+      content: "";
+      position: absolute;
+      display: none;
+  
+  .label-text
+    padding-left: 34px;
+    white-space: nowrap;
```
bobheadxi: can we get newlines on the end of files? 😛 

https://marketplace.visualstudio.com/items?itemName=jdforsythe.add-new-line-to-files

web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -0,0 +1,54 @@
...
+    opacity: 0;
+    cursor: pointer;
+
+    &:checked ~ .radio-img
+      background-image: url('../assets/radio/selected.svg');
+      &:after
+        display: block;
+
+    &:disabled
+      cursor: not-allowed;
```
bobheadxi: Not a big fan of manipulating cursors - is the disabled style and unclickableness of an element not enough feedback?

Maybe @sherryyx could weigh in? (this is what causes the big `x` on the mouse that the demo gif shows)

chamkank: It was just something I added to make it extra clear that the input is disabled, @sherryyx it can be removed if you don't want it from a UX perspective.

chamkank:
I'll make the changes from #39 in a separate PR for both the checkbox & radio button components.

chamkank:
( Closes #38 )

PR #41, opened 2018-08-09
Web/navbar hide

### LINKED ISSUE #35: Make navbar fade and reappear as user scrolls

bobheadxi:
Closes #35

- Reworked `navbar` styles based on #39 
- Reworked `navbar` logic
- Added animation for scroll
- Updated demo

navbardemo-2

## :thought_balloon: Notes

Targeted at #39 for now for a clearer diff, do not merge! Will point to `master` once that is in.

```make web``` go to `/ui_demo`

## To Do

- [x] Update tests 😢

### CODE REVIEW

File: web/components/navbar/index.js [hunk 1]
```diff
@@ -23,43 +20,95 @@ const getButton = (buttonType) => {
...
-            <div>{getButton(buttonType)}</div>
-          </div>
-        </nav>
-      );
-    default:
-      return (
-        <nav>
-          <div>
-            <div>{getLogo()}</div>
+class Navbar extends React.Component {
```
mingyokim: I believe you need to add a `componentShouldUpdate` lifecycle method to update the state when new props are passed into this component.

Right now the component is not updating based on incoming props. Reproduction steps:
* Make sure you're logged out
* Go to /
* Refresh and it should have logo, links, and login button
* Click "login" button
* The navbar still has logo, links, and login button
* Refresh
* The navbar is updated to just have logo

bobheadxi: Can't repro anymore after f66d3b0 - let me know if it works when you get the chance!

mingyokim: Yeah I can't reproduce it anymore. But wouldn't it be better to have `displayType` and `buttonType` in the `state`? The only reason why this component re-renders properly is because its container is re-rendering itself. If for some reason the container does not re-render itself, but the props change, then this component wouldn't re-render.

web/components/navbar/index.js
```diff
@@ -23,43 +20,93 @@ const getButton = (buttonType) => {
...
+  }
+
+  handleScroll = () => {
+    // Calculate scroll direction
+    const { lastY } = this.state;
+    const scrollingDown = (window.scrollY - lastY) >= 0;
+
+    // Calculate position
+    const offset = window.pageYOffset || document.documentElement.scrollTop;
+    const atTop = offset < 96;
```
mingyokim: Can we store `96` into some kind of variable?

web/components/navbar/index.js
```diff
@@ -23,43 +20,93 @@ const getButton = (buttonType) => {
...
+
+    // Calculate transparency
+    const transparent = offset < 256;
```
mingyokim: same here

web/components/navbar/index.js
```diff
@@ -23,43 +20,93 @@ const getButton = (buttonType) => {
...
+      hidden: (!atTop && scrollingDown),
+      lastY: window.scrollY,
+    });
+  }
+
+  render() {
+    const { hidden, transparent } = this.state;
+    const { displayType, buttonType } = this.props;
+    const button = getButton(buttonType);
+    const links = [
```
mingyokim: cleaann but can we name this something like `linksAndButton`? it doesn't just have links

web/styles/spacing.sass
```diff
@@ -1,11 +1,78 @@
...
+    &-text-inputs
+      > .text-input
+        &:first-of-type
+          margin-bottom: $spacing-xl
+        &:last-of-type
+          margin-bottom: $spacing-mega
+
+  // repeating horizontal elements
+  &-horizontal
+    &-divs
```
mingyokim: nice

web/components/navbar/Navbar.sass
```diff
@@ -1,43 +1,21 @@
...
     font-family: Apercu Pro
     font-size: 18px
-    color: $primary
-  a:visited
-    color: $primary
 
-  a:hover
-    color: #00D5D5
   img
+    margin-top: 3px
```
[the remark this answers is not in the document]
bobheadxi: This is a hack because the logo looked slightly off :(

web/components/navbar/index.js
bobheadxi: Sure - 96 is in the CSS as navbar height, but I’ll replicate it here

web/components/navbar/index.js
bobheadxi: True!

web/components/navbar/Navbar.sass
```diff
@@ -1,43 +1,21 @@
 nav
-  display: flex
+  z-index: 500
```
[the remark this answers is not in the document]
bobheadxi: This keeps the navbar above all other elements, so long as we avoid using z-index elsewhere

web/components/navbar/index.js
bobheadxi: Hm, I’ll have a closer look at this tonight

bobheadxi: @mingyokim sorry, I'm not sure I understand - doesn't a prop change force the container to re-render? From what I understand, `displayType` and `buttonType` are only manipulated from here, and the navbar itself does not change them:

https://github.com/nwhacks/nwhacks2019/blob/master/web/containers/navbar/index.js#L57

It also seems like setting `props` that is somewhat bad practice (https://stackoverflow.com/a/47341539), but there also seems to be some other APIs for doing this kind of stuff (https://stackoverflow.com/a/49839457, https://stackoverflow.com/a/32414771). From what I understand, this type of stuff should just be kept in `props` (https://stackoverflow.com/a/40330195). I could be missing something though...

web/components/navbar/Navbar.sass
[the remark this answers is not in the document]
bobheadxi: 4px looks fine too though - I'll make this change

web/components/navbar/index.js
mingyokim: @bobheadxi There are two types of React components: a "stateful" component (the one defined by a class), and a stateless component (the one defined by a function). A prop change does not force the stateful component to re-render, only a state change does (http://lucybain.com/blog/2017/react-js-when-to-rerender/). (aside: prop change DOES cause a component to re-render if it's a stateless component)

> doesn't a prop change force the container to re-render?

Did you mean `component` instead of `container`? Based on the information above, a prop change does not force the component to re-render.

I would agree that copying `props` that never change into `state` is a bad practice. However, in our case, `buttonType` and `displayType` do change, although those changes aren't triggered by the component itself, but by its container. So I believe it would be best to set those as `state` so that whenever the container passes in different values for `buttonType` and `displayType`, the component is _guaranteed_ to re-render, and not relying on the container to re-render for this component to be re-rendered.

Technically, you could prove that the container always re-renders whenever the `buttonType` and `displayType` changes. If you can, great, we can leave it as is, along with some comments (since it's not plain obvious). But I'm pretty sure that guaranteeing that the component re-renders itself whenever `buttonType` and `displayType` change by setting them as `state` instead of coming up with a proof is the path of least resistance.

bobheadxi: https://stackoverflow.com/a/51525302

bobheadxi: > By default, when your component's state or props change, your component will re-render.

https://stackoverflow.com/a/35004739

bobheadxi: > setState will trigger componentUdpate -> componentWillUpdate -> render. **props change will trigger componentWillReceiveProps before this chain.**

Prop changes will end up calling `render()`.

> Updating state by setState will trigger the render function (re-render). **And props also trigger render as well.**

https://stackoverflow.com/a/45654488

bobheadxi: > A prop change does not force the stateful component to re-render, only a state change does 

This seems ridiculous if true (and I really doubt it is - the article doesn't seem to suggest it either, plus see the screenshot from earlier) - a stateful component's `props` are exactly the same as a stateless component's `props`... it's just that one type has a state that can enact additional re-renders

> Conceptually, components are like JavaScript functions. They accept arbitrary inputs (called “props”) and return React elements describing what should appear on the screen.

The official react docs suggest no difference in how the two types handles props, other than additional lifecycle hookshttps://reactjs.org/docs/components-and-props.html

bobheadxi: Officially, react only distinguishes between class and function (potentially stateful and stateless respectively)

https://reactjs.org/docs/components-and-props.html#functional-and-class-components

mingyokim: so props do cause a render on stateful components? oops i was mistaken, it's all good then

bobheadxi: Similar question: https://www.reddit.com/r/reactjs/comments/7tn79g/state_and_componentwillreceiveprops/

```js
class Foo extends Component {
  state = { value: this.props.value }

  // If the props change, update state
  componentWillReceiveProps (nextProps) {
    if (nextProps.value !== this.props.value) {
      this.setState({ value: nextProps.value })
    }  
  } 

  render () {
    // render based on state.value
  }
}
```

Top response:

> Why you don't render props directly?

mingyokim: > An update can be caused by changes to props or state.

https://reactjs.org/docs/react-component.html

Okay, so clearly change in props cause a re-render, `reactjs` says so. My bad lol

If this is the case, definitely use props rather than state because the component is not modifying the `buttonType` or `displayType`, they're just displaying it. 

Approving

mingyokim: I'll just do one last sanity check

PR #51, opened 2018-08-12
Web/refactor/checkbox+radio

### LINKED ISSUE #48: Refactor styles for radio button & checkbox
https://github.com/nwhacks/nwhacks2019/pull/38#issuecomment-411847818

chamkank:
:tickets: **Ticket(s)**: Closes #48 

---

1. Bug fixes for disabled ghost label and `checked` attribute.
2. Added classes to `utils.sass` and `typography.sass`.
4. `RadioButton`, `RadioGroup`, `Checkbox`, `CheckboxGroup`
    - Added `className` prop
    - Added `sharedClassName` prop to `RadioGroup` and `CheckboxGroup` (see notes)
    - Refactored styles
    - Made `CheckboxGroup` and `RadioGroup` use `.flex` and `.row` by default (see notes)
5. Added `utils.js` file for shared functions between components.
## :thought_balloon: Notes

1. The className convention that we decided on previously conflicts with the convention of shared props for `RadioGroup` and `CheckboxGroup`. For that reason I added `sharedClassName` which is applied to all children. So `className` is used just applied to the container.
2. I figured that we would most likely be using `.flex` and `.row` every time we use these components so I made that the default style. 
3. Ignore commits 89c4c26 to 3a16ad2 (reverted changes so the diff is fine). 

1. `make web` and go to `localhost:8080/ui_demo`.
2. Resize the window and see the checkbox and radio button components flex. Note that `dir-col` is applied to `RadioGroup` through `className`.


web/components/input/buttons/CheckboxGroup/Checkbox.sass
```diff
@@ -1,54 +1,49 @@
...
+      content: ""
+      position: absolute
+      display: none
   
   .label-text
-    padding-left: 34px;
-    white-space: nowrap;
\ No newline at end of file
+    cursor: pointer
+    padding-left: 34px
```
mingyokim: spacing variable

web/components/input/buttons/CheckboxGroup/index.js
```diff
@@ -57,6 +57,20 @@ CheckboxGroup.propTypes = {
...
+
+      // see Checkbox.propTypes
+      name,
+      value,
+      defaultChecked,
+      disabled,
+      onChange,
+
+      // passed down to all children (but not applied to container)
+      sharedClassName: PropTypes.string,
```
mingyokim: shouldn't this be a proptype of checkbox?

web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -1,54 +1,49 @@
...
-  color: $primary;
-  position: relative;
-  margin-bottom: 8px;
-  cursor: pointer;
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
-  user-select: none;
+  @extend .user-select-none
+  margin-left: 20px
```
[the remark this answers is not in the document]
bobheadxi: We should have a spacing class for this as well. I don't think the `radio-group` and `radio-button` classes need to exist

web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -1,54 +1,49 @@
+.radio-group
+  @extend .flex, .wrap
```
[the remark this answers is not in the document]
bobheadxi: This is used here as well: https://github.com/nwhacks/nwhacks2019/blob/master/web/styles/decoration.sass#L5

That being said, this would be much better declared inline in the JSX

web/components/input/buttons/CheckboxGroup/Checkbox.sass
bobheadxi: Spacing class, actually - most of these properties can be declared using inline classes.

web/components/input/buttons/CheckboxGroup/Checkbox.sass
```diff
@@ -1,54 +1,49 @@
...
-    left: 0;
-    height: 24px;
-    width: 24px;
+    position: absolute
+    top: 0
+    left: 0
+    height: 24px
+    width: 24px
     background-image: url('../assets/checkbox/unchecked.svg')
+    cursor: pointer
```
bobheadxi: make this a `clickable` class

web/components/input/buttons/CheckboxGroup/Checkbox.sass
```diff
@@ -1,54 +1,49 @@
...
       &~ .checkmark
-        cursor: not-allowed;
-        opacity: 0.4;
+        cursor: not-allowed
+        opacity: 0.4
         &~ .label-text
-          cursor: not-allowed;
-          opacity: 0.4;
+          cursor: not-allowed
+          opacity: 0.4
```
bobheadxi: Class, we don't need repeated styles, and if we are using `cursor: not-allowed` etc they should be part of a `unclickable` class that all our buttons can use.

web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -1,54 +1,49 @@
...
-    background-image: url('../assets/radio/hover.svg');
+    background-image: url('../assets/radio/hover.svg')
 
   input[type="radio"]
-    position: absolute;
-    opacity: 0;
-    cursor: pointer;
+    position: absolute
+    opacity: 0
+    cursor: pointer
```
bobheadxi: class, this is repeated.

web/styles/utils.sass
```diff
@@ -82,3 +82,12 @@
   &.ai
     &-center
       align-items: center
+
+  // CSS Basic User Interface helper classes
+  .user-select
```
bobheadxi: are you sure this shouldnt be `&.user-select` or unindented?

web/components/utils.js
```diff
@@ -0,0 +1,5 @@
+// contains helper functions shared by component library
+
+const concatClassName = (baseClassName, extraClassName = undefined) => (extraClassName ? [baseClassName, extraClassName].join(' ') : baseClassName);
```
bobheadxi: this seems like a pretty useless function - why not just use a template string?

```js
className={`defaultClass ${className}`}
```

avoids the unnecessary import

web/components/input/buttons/CheckboxGroup/index.js
bobheadxi: I'm not sure I understand how this works, or why it's needed - I can think of two situations where children are generated:

1. User provides children to class. In this case, sharedClassName is useless, because you can just attach className to the provided children
2. Component generates children. In this case, isn't sharedClassName also useless, because you don't want the user to influence how these children are generated, because they are tied to the parent component? Why can't the parent just provide the desired classNames?

can you provide an example/point us to one? a quick cmd-f doesn't find me any usage of this prop

web/styles/utils.sass
```diff
@@ -82,3 +82,12 @@
...
+    &-none
+      -webkit-user-select: none
+      -moz-user-select: none
+      -ms-user-select: none
+      user-select: none
+
```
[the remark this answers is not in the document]
chamkank: Fixed

web/components/input/buttons/RadioGroup/RadioButton.sass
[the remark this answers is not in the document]
chamkank: Fixed

web/components/input/buttons/RadioGroup/RadioButton.sass
```diff
@@ -1,54 +1,49 @@
...
 
     &:disabled
-      cursor: not-allowed;
-      opacity: 0.4;
+      cursor: not-allowed
       &~ .radio-img
-        cursor: not-allowed;
-        opacity: 0.4;
+        cursor: not-allowed
+        opacity: 0.4
```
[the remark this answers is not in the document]
chamkank: Fixed, declared an opacity class in `utils.sass`

web/components/input/buttons/RadioGroup/RadioButton.sass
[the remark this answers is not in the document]
chamkank: Fixed

web/components/input/buttons/CheckboxGroup/Checkbox.sass
chamkank: Fixed

web/components/input/buttons/CheckboxGroup/index.js
chamkank: In `RadioGroup` and `CheckboxGroup`, `className` is intended to apply class names to that container, and so it's not passed down to the children (unlike `name`, `disabled`, `selected/defaultChecked`, and `onChange` props). However, `sharedClassName` is passed down to the children.

So instead of this...
```
<RadioGroup name="test_radio_group">
  <RadioButton label="Label 1" value="value1" className="margin-right-m" />
  <RadioButton label="Label 2" value="value2" className="margin-right-m" />
  <RadioButton label="Label 3" value="value3" className="margin-right-m" />
</RadioGroup>
```
... we can do this:
```
<RadioGroup name="test_radio_group" sharedClassName="margin-right-m">
  <RadioButton label="Label 1" value="value1" />
  <RadioButton label="Label 2" value="value2" />
  <RadioButton label="Label 3" value="value3" />
</RadioGroup>
```

This is also the reason why `sharedClassName` isn't a prop of the children components.

In retrospect, the convenience is probably not worth the confusion (if you define `className` for a `RadioButton`/`Checkbox`, it will override `sharedClassName` instead of adding to it). I can revert it back to the state where class names are not passed down to the children (similar to the `value` prop) and must be explicitly defined for each child. Let me know which one you guys like more.

web/components/input/buttons/CheckboxGroup/Checkbox.sass
chamkank: Fixed

web/components/input/buttons/CheckboxGroup/index.js
bobheadxi: Yeah, I think it's probably best to be more explicit and let className be defined for each child. If you find it verbose, a simple `map` will do the trick:

```js
<RadioGroup name="test_radio_group">
  {radioButtons.map(b => (<RadioButton
    key={b.label} 
    label={b.label}
    value={b.value} 
    className="margin-right-m" />)}
</RadioGroup>
```

web/styles/typography.sass
```diff
@@ -47,3 +47,10 @@ h2
 
 h5
   font-size: 18px
+
+label
+  font-family: Apercu Pro
+  font-size: 18px
+  font-weight: bold
+  line-height: normal
+  color: $primary
```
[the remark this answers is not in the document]
chamkank: Fixed, also the ghost label bug was fixed by d83e3fb078f9947a82c2469903b4bce6a7b5be14, the 0.4 overrided the zero opacity of the input field. The [`checked` attribute bug](https://i.imgur.com/7EsSpli.mp4) was fixed with 32ea45b14449964ddd7ddc1c4321b7f26723e288

web/components/utils.js
chamkank: ```className={`defaultClass ${className || sharedClassName}`}``` resulted in `undefined` being in the class name when neither variables were defined (even adding `|| ""` to the end of the ternary didn't work). I'll go ahead and remove this since I'll be removing `sharedClassName` altogether.

bobheadxi: Since convention is to destructure prop args straight off the bat, you can provide defaults:

```js
const myComponent = ({ className = '' }) => (< ... />)
```

I might have forgotten to do this though, I'll add it if I see it

mingyokim: ooh good catch, we should handle `className` being undefined

mingyokim: we could probably enforce this by setting `className` as required with `propTypes`

chamkank:
I don't think my rebase worked as intended... :face_with_head_bandage:

bobheadxi:
oh dear hahah

maybe try `git reset --soft HEAD~83`, then commit changes back one by one, or we can squash this whole deal using GitHub?

chamkank:
Yikes the first one sounds like a headache, guess I'll do the mega squash when merging. :man_dancing:

bobheadxi:
thanks @chamkank !

chamkank:
Thanks for all the feedback guys!

PR #56, opened 2018-08-17
Homepage + scaling

### LINKED ISSUE #34: Implement homepage main sections

bobheadxi:
:tickets: **Ticket(s)**: Closes #34

---

First few rows of the homepage + lots of new scaling stuff. I feel stuff is finicky... can't seem to get the fonts to quite match the Figma... SVG scaling needs improvmeent and very low widths

image
image
image

## :thought_balloon: Notes

Won't be including FAQ or Stories in this PR.

```
make web
```


web/components/home/index.sass
```diff
@@ -0,0 +1,15 @@
+.homepage
+  background: $gradient-purple
+
+  &-email-registration
+    > .text-input
+      width: 300px
```
mingyokim: this should probably be `width: 100%` and `max-width: 300px`?

web/components/home/Home.js
```diff
@@ -0,0 +1,106 @@
...
+
+const getImageDiv = (alt, src) => <img alt={alt} src={src} />;
+
+const rowStyle = `
+  flex jc-between ai-center dir-row
+  pad-sides-ultra pad-bottom-giga scale-pad-sides-tablet`;
+
+const Home = () => (
+  <div className="homepage overflow-hidden">
+    <div className="flex jc-between ai-center pad-top-ultra pad-sides-ultra scale-pad-sides-tablet homepage-scale-top">
```
mingyokim: actually really digging these utility classes, makes css so much cleaner omg

web/components/home/index.sass
bobheadxi: omg genius

web/components/home/Home.js
```diff
@@ -0,0 +1,111 @@
...
+        {getImageDiv('sun', sun)}
+      </div>
+    </div>
+
+    <div className="homepage-skyline fill-width">
+      {getImageDiv('skyline', skyline)}
+    </div>
+
+    <div className={`${rowStyle} scale-row-desktop`}>
+      <div>
```
mingyokim: Can we make each of these take 50% of the width? Kinda looks weird in 4k, the text occupies a lot more space than the image and it looks really thinned out

homepage-4k

web/styles/breakpoints.sass
```diff
@@ -7,3 +7,61 @@ $laptop: 1080px;
...
 $mobile: 480px;
+
+//////////////////////////
+//   SCALING CLASSES    //
+//////////////////////////
+
+.scale
+  &-row
+    &-desktop
+      @media screen and (max-width: $desktop)
```
mingyokim: hm why `$desktop`? Shouldn't it be `$laptop`? 

Also this might be out of scope but why is `$desktop` `1300px`? Shouldn't `$laptop` be `1024px`, and `$2k` be `1440px` and `$4k` be `2560px`?

web/styles/breakpoints.sass
```diff
@@ -7,3 +7,61 @@ $laptop: 1080px;
...
+        @media screen and (max-width: $desktop)
+          flex-direction: column-reverse !important
+
+  &-pad
+    &-sides
+      &-desktop
+        @media screen and (max-width: $desktop)
+          padding-right: $spacing-s !important
+          padding-left: $spacing-s !important
+        &-none
```
mingyokim: shouldn't this go under `-tablet` since it's a break point for `$tablet`?

web/components/home/Home.js
```diff
@@ -0,0 +1,111 @@
...
+        <p>
+          Come make things and break things, and then make them cooler. You&apos;ll
+          never be short on inspiration when you’re surrounded by 600 of the
+          brightest minds in the Pacific Northwest. All you need to bring is an
+          open mind and an insatiable desire to learn; we’ll take care of the
+          rest. After all, we&apos;re western Canada’s largest hackathon — we make
+          the west coast the best coast.
+        </p>
+      </div>
+      <div className="pad-left-giga pad-ends-l scale-pad-sides-desktop-none scale-jc-center">
```
mingyokim: where's `scale-jc-center` defined?

web/components/home/index.sass
```diff
@@ -0,0 +1,21 @@
+.homepage
+  background: $gradient-purple
+
+  &-email-registration
+    @media screen and (max-width: 300px)
```
mingyokim: Can we follow google chrome standard and use `$mobile-s` for `320px`?

web/components/input/buttons/PrimaryButton/PrimaryButton.sass
```diff
@@ -4,7 +4,6 @@ button.primary
   color: $white
 
   border-radius: 4px
-  min-width: 80px
```
mingyokim: why are we removing `min-width` here?

web/components/home/Home.js
```diff
@@ -0,0 +1,111 @@
...
+        <p className="primary">
+          Western Canada&apos;s largest hackathon<br />
+          January 26-27, 2019 @ the University of British Columbia
+        </p>
+        <p className="secondary">
+          Get notified when registration opens!
+        </p>
+        <div className="homepage-email-registration flex ai-end">
+          <TextInput
+            placeholder="hacker@email.com" />
```
mingyokim: email-error

Probably need to define `onChange` function here? And save the email as a state of this component (this could be addressed in another PR)?

bobheadxi: righto ill add a placeholder for now and open a ticket for hooking it up with actual functionality

web/styles/spacing.sass
```diff
@@ -11,6 +11,8 @@ $spacing-xl: 40px;
 $spacing-xxl: 48px;
 $spacing-mega: 56px;
 $spacing-giga: 64px;
+$spacing-super: 96px;
```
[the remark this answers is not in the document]
bobheadxi: OH thats where you got it from ive just been making up prefixes

web/components/input/buttons/PrimaryButton/PrimaryButton.sass
bobheadxi: this sets to min-width to text + padding rather than a flat `80px` which causes the wonky clipping on smaller screens

web/components/home/index.sass
bobheadxi: the 300px is from figma - should I change it anyway?

web/styles/breakpoints.sass
bobheadxi: I sniped the breakpoints from https://github.com/ubclaunchpad/ubclaunchpad.github.io/blob/master/css/scss/utils/_breakpoints.scss because imo they are quite reasonable

I think `$2k` etc arent very good breakpoints - very few displays render websites at native resolution (macOS for example upscales by default, and i think a lot windows is configured to do the same nowadays but maybe @chamkank can pitch in on that)

bobheadxi: I chose to do desktop (1300) vs laptop (1080) here because after fiddling around with screen widths for a while i found it was better to break the homepage on this

web/components/home/Home.js
bobheadxi: unused style! will remove oops

web/styles/breakpoints.sass
bobheadxi: ye it should
so many styles to keep track of rip

web/components/home/Home.js
bobheadxi: saw you other comment before this but how do you define 4k and whatnot

but yes ill fix this good catch

web/styles/breakpoints.sass
bobheadxi: tl;dr: I will fiddle around with changing the breakpoints, but imo `desktop laptop tablet phablet mobile` are good points to break at because those are the ones where things can **really look bad**, and we should focus on making accommodating those, since if it looks good at `1920` chances are itll look okay at most reasonable screen sizes past that and we dont need to make special scaling classes for them

web/components/home/Home.js
bobheadxi: @mingyokim actually it might be best to leave this as is - if we scale the image up to 50% of the screen, we get massive images (to preserve aspect ratio) and tiny text

web/styles/breakpoints.sass
bobheadxi: actually no imma leave `desktop` as is - i think `1300` is a good point to define "alright, at this width, we switch into desktop layout", `1920` is way to big to be useful for breakpoints

mingyokim: Yeah, I agree that we don't need breakpoints past the standard non-mobile resolution (2k and 4k are probably excessive).

I'm not sure why we need both `desktop` and `laptop`... don't most displays look the same between a desktop and laptop? and i grabbed the numbers i came up with for $2k etc from chrome responsive mode, but i don't understand where they got those pixels from so scratch that. But I think we should keep it consistent with the standard resolution like `1080p (1920 x 1080)`.

so I'd say we just have one break point, 1920px, for non-mobile resolution and keep the lower ones the same?

web/components/home/index.sass
mingyokim: yeah, small phones like "iPhone SE" have a width of 320px

web/components/input/buttons/PrimaryButton/PrimaryButton.sass
mingyokim: interesting, good to know!

web/components/navbar/index.js
```diff
@@ -86,20 +86,20 @@ class Navbar extends React.Component {
         navbarRight = navbarRightElements.map(l => (
           <div
             key={key += 1}
-            className="flex ai-center margin-sides-l">
+            className="flex ai-center margin-sides-l scale-margin-sides-tablet scale-hide-phablet">
```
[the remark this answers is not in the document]
mingyokim: PR: #62

web/styles/breakpoints.sass
bobheadxi: The problem with using `1920` as a breakpoint is that I dont really see a case where something looks good from like `1000` to `1920` and then abruptly looks awful after

I think you might be looking at the breakpoints a bit differently - these are intended to be the "line" between "classes" of display types, where each class of display type might need some kind of CSS adjustment. `laptop` here is useful as a point to use where a slightly smaller screen is juuuusssst to small to comfortably display something... in this case, the sun at the top banner. A `1920` is not useful here, because theres not much that will change *past* `1920` where CSS adjustments need to be made, whereas its too wide to say this is the point where something becomes a desktop... I think `1300`, from what ive noticed fiddling with this PR, is a comfortable entry into the `desktop layout` zone

mingyokim: oh i see, i understand now. Yeah I was thinking about it differently, but that makes sense.

web/components/home/Home.js
bobheadxi: I'm improved this a little with 1860d39

It doesnt look bad, but im not sure whats causing the borders to kinda not respect the parent div:

image

i hate css

web/styles/breakpoints.sass
mingyokim: this still needs to go under `-tablet`

web/components/navbar/index.js
```diff
@@ -66,11 +68,11 @@ class Navbar extends React.Component {
...
-      <Link to="/"><b>About</b></Link>,
-      <Link to="/"><b>Stories</b></Link>,
-      <Link to="/"><b>FAQ</b></Link>,
-      <Link to="/"><b>Sponsors</b></Link>,
-      <Link to="/"><b>2018</b></Link>,
+      <Link to={{ pathname: '/', hash: SECTION.ABOUT }}><b>About</b></Link>,
+      <Link to={{ pathname: '/', hash: SECTION.STORIES }}><b>Stories</b></Link>,
+      <Link to={{ pathname: '/', hash: SECTION.FAQ }}><b>FAQ</b></Link>,
+      <Link to={{ pathname: '/', hash: SECTION.SPONSORS }}><b>Sponsors</b></Link>,
+      <a href="http://nwhacks.github.io/nwhacks2018_static"><b>2018</b></a>,
```
mingyokim: Can we add `target=_blank` to open this page in a new tab ? Since it’s an external link

web/components/navbar/index.js
```diff
@@ -79,35 +81,36 @@ class Navbar extends React.Component {
...
           </div>
         ));
+        navbarRight.push(<div key={key += 1} className={LINK_CLASS}>{button}</div>);
         break;
       default:
         navbarRight = linkElements.concat([button]).map(l => (
           <div
             key={key += 1}
-            className="flex ai-center margin-sides-l">
+            className={`${LINK_CLASS} scale-hide-phablet`}>
```
mingyokim: Wouldn’t this hide the button if the viewport is phablet? We only want to hide the links here

PR #59, opened 2018-08-18
Web/application/hacker/sign up

### LINKED ISSUE #55: Hacker Application Sign Up UI
Implement UI for hacker application sign up page

mingyokim:
:tickets: **Ticket(s)**: Closes #55 

---

sign-up-desktop
sign-up-mobile

* Dynamically enable/disable the "Next button" on hacker application
* Implement the UI for sign up page
* Update `TextInput` and `PasswordInput` props to be able to take in more props and customize them more.

## :thought_balloon: Notes

N/A

- [ ] Passwords clear when refreshing (we probably don't want to store passwords in local storage for security reasons)
- [ ] Next button is disabled unless passwords match and are not empty
- [ ] Updates to email are persisted


web/components/application/hacker/Hacker.js
```diff
@@ -32,19 +33,34 @@ const HackerApplication = (props) => {
...
+    enableNextButton,
+    disableNextButton,
   } = props;
 
   const indexToPage = {
     0: (
       <PageOne
         hackerApplication={hackerApplication}
         onHackerApplicationChange={onHackerApplicationChange}
+        enableNextButton={enableNextButton}
```
bobheadxi: this seems brittle... why not a single callback like `fieldsAreValid()` to indicate that the page is next-able?

web/styles/animations.sass
```diff
@@ -34,6 +34,7 @@ $transition-time-long: 1.5s
     width: 100%
     z-index: -100
     -webkit-transition: $transition
-  &:hover
-     &:before
-       opacity: 1 
+  @media (hover:hover)
```
bobheadxi: does this have something to do with making this work with images?

web/styles/utils.sass
```diff
@@ -58,6 +58,16 @@
...
 
+  // algin-self
+  &.as
+    &-end
+      align-self: flex-end
+
+  // align-content
+  &.ac
+    &-end
+      align-content: flex-end
```
bobheadxi: 💪

web/styles/animations.sass
mingyokim: nope http://www.javascriptkit.com/dhtmltutors/sticky-hover-issue-solutions.shtml
I think in mobile when you click something it stays in `hover` mode and this is one way to deal with it

web/components/application/hacker/Hacker.js
mingyokim: This is not to "indicate" whether or not the page is "next-able". This is an action/function to actually enable the next button.

bobheadxi: i might be missing something, but isnt enabling the next button just a roundabout way of indicating that this page is `next-able` (if "next-able", enable my button - if not "next-able", disable my button)?

mingyokim: Oh are you asking why we need both `enableNextButton` and `disableNextButton` instead of just `enableNextButton`/`fieldsAreValid`?

If that's what you're asking, we also need to be able to disable the next button when the form in the page is no longer valid.

bobheadxi: but if you have:

```js
fieldsAreValid(valid? bool)
```

child:

```js
// insert destructuring lol
const valid = this.state.isValid; // true or false
fieldsAreValid(valid)
```

parent:

```js
fieldsAreValid(valid => if (valid) { enableButton(); } else { disableButton(); }
// this is definitely not valid syntax
```

mingyokim: oh yeah that's doable. but why do you exactly prefer it this way?

bobheadxi: less things to pass around, and imo having two functions for what is essentially passing back a boolean state might be a bit much

bobheadxi:
things I ran into:

- we should include some *super basic* email validation, like check for the presence of `@` and two components on `email.split('@')`, and maybe at least one `.` in the second half
- warnings:
image
- when inputting passwords for the first time, the `next` button enables and disables correctly as a type the confirmation, but the error does not show up until I deselect the text box. When I do that, error updates correctly as I type. felt a little weird - is this intentional?

out of scope for PR but we might want to automate the redux store clear at some point, since i imagine it might cause issues if we update the site after it goes public... do you think thats possible?

mingyokim:
@bobheadxi 
* Hmmm I COULD add some basic regex for email validation, but ultimately firestore does email validation for us when we make a request to create an account and we would display the error accordingly - so knowing this, is there really a value in doing that?
* That should be fixed by #60 
* Yeah, that was intentional behaviour. I guess this is a question of behaviour, so will get @sherryyx 's approval (unless you strongly feel like this isn't the right design, then you can chip in too)
* Hmmm yeah we have functions that clear redux store but it's hard to tell when to execute them? I can look into it though.

bobheadxi:
> but ultimately firestore does email validation for us when we make a request to create an account and we would display the error accordingly - so knowing this, is there really a value in doing that?

this would only happen at the end, no? its probably a better experience to have this show up live a la the mismatching passwords

> Yeah, that was intentional behaviour

mmk thats fine, i think the best experience ive had with this type of stuff is where a check runs after x seconds of not typing (which then displays the error), but probs a bit of a hassle

> Hmmm yeah we have functions that clear redux store but it's hard to tell when to trigger them

ya what i was thinking too

mingyokim:
@bobheadxi do you mind if we revisit email validation later? It could always be implemented in a separate PR. And do you have any other comments other than that?

PR #60, opened 2018-08-18
Web/refactor/input usage

### LINKED ISSUE #57: Refactor input usage
To make our input components use implicit labels (`<label>` wraps `<input>` and input uses `name` field instead of `id`).

chamkank:
:tickets: **Ticket(s)**: Closes #57 

---

Just a small refactor of our input components to use the same conventions before we move forward with the rest of the components. Here are the changes:
1. Used implicit labels instead of explicit labels so we don't need to have a unique `id` for each component (just the input `name` instead).
2. Added `type` field to `PrimaryButton` and `SecondaryButton` (see notes).
2. Updated usage of input components.
3. Updated tests.

## :thought_balloon: Notes

~~For `PrimaryButton`, `SecondaryButton`, and `ProgressGroup`, the `name` is optional. For the buttons, the button type becomes `submit` if you provide a value for `name`. If you don't provide a `name` value, the button type defaults to `button`.~~ You can also explicitly define the `type` (to [`submit`, `button`, or  `reset`](https://www.w3schools.com/tags/att_button_type.asp) with the default being `button`) of `PrimaryButton` and `SecondaryButton`.


web/components/application/hacker/Hacker.test.js
```diff
@@ -246,7 +246,7 @@ describe('Hacker component', () => {
         });
 
         it('page one is rendered', () => {
-          expect(wrapper.find('div#hacker-application-page-1')).toHaveLength(1);
+          expect(wrapper.find('form#hacker-application-page-1')).toHaveLength(1);
```
bobheadxi: @mingyokim an example of the potentially-overly-specific-tests i talked about the other day

mingyokim: yeah, i get it now. i'll try to avoid these patterns next time

web/components/input/buttons/PrimaryButton/index.js
```diff
@@ -1,12 +1,13 @@
...
-const PrimaryButton = ({ text, onClick, disabled, className }) => (
+const PrimaryButton = ({ text, onClick, disabled, className, name, type }) => (
   <button
     onClick={onClick}
-    type="submit"
+    type={type ? type : (name ? "submit" : "button")}
     disabled={disabled}
-    className={`primary ${className}`}>
+    className={`primary ${className}`}
+    form={name}>
```
mingyokim: do we really need ` form` here? It's not submitting even when the type is `submit` if `form` is defined

chamkank: Oops, `form` was supposed to equal the form-id but I made it the form-name. Fixed by fd5ad1c75187ec52ec1fd7a56af88ebb0f7eccdf. The commit also removes this attribute altogether since it's only really useful in the extreme case where you want to associate a button with a form when it's outside the `<form>` tags.

web/.eslintrc.json
```diff
@@ -22,6 +22,7 @@
     "react/prefer-stateless-function": 0,
     "react/no-did-mount-set-state": 0,
     "react/jsx-one-expression-per-line": 0,
+    "react/button-has-type": 1,
```
chamkank: Due to https://github.com/yannickcr/eslint-plugin-react/issues/1846 and https://github.com/yannickcr/eslint-plugin-react/issues/1555

chamkank:
Thanks guys!

PR #90, opened 2018-09-04
set container width to 100%

mingyokim:

---

Sets container width to 100% - on mobile we want the container to occupy full width with given padding

Make sure on mobile view the padding of the container in home page matches the design


web/styles/utils.sass
```diff
@@ -10,7 +10,7 @@
     height: 100%
 
 .container-width
-  width: 90%
+  width: 100%
```
bobheadxi: doesn't this default to 100% if we remove `width`?

mingyokim: no it defaults to the width of the contents inside the container. We want the containers to occupy 100% of the width even if the content don't necessarily fill up 100% of the width.

Without `width: 100%`:
no-width

With `width: 100%`:
width-100

PR #98, opened 2018-09-05
add maxLength to text input

### LINKED ISSUE #58: Set character limit for text inputs
Don't want to allow users to put unpredictable load on firestore by inputting a very large string

mingyokim:
:tickets: **Ticket(s)**: Closes #58 

---

Sets maximum character limit for text inputs (set default to 100 for now)

Make sure you can't type more than 100 characters?


chamkank:
Shouldn't this be defined on a case-by-case basis with the default being no limit?

chamkank:
I definitely agree that there should be a limit. What I meant by my last comment is that the max length should be required and explicitly defined when it's used so that when you're reading the code at the level of the form you can clearly understand the limit for each of the fields. i.e. the detail shouldn't be hidden

```
<TextInput id="fname" maxCharNumber=100 .../>
<TextInput id="lname" maxCharNumber=100  .../>
<TextInput id="email" maxCharNumber=255 .../>
```

Otherwise if you have a bunch of text inputs where only the inputs that have a different maxLength from 100 have the field explicitly defined, you might think "oh, only some of these fields have a character limit" unless you've already looked at the defaultProps for the TextInput component. Seems more natural to think that `fname` and `lname` have no character limits in the below example.

```
<TextInput id="fname" .../>
<TextInput id="lname" .../>
<TextInput id="email" maxCharNumber=255 .../>
```

I guess we could also just leave the default 100 limit but explicitly define it like in the first example when we use it as a best practice. But even then I feel it would be less misleading to have no limit by default instead of an arbitrary default limit. :thinking: Let me know what you think

mingyokim:
@chamkank yes i agree with your alternative - my intention was to never having to specify the maximum number of characters except in extraneous cases

web/components/input/text/TextInput/TextInput.js
```diff
@@ -50,7 +50,7 @@ const TextInput = ({
 TextInput.defaultProps = {
   className: '',
   sideLinkOnClick: () => {},
-  maxCharNumber: 100,
+  maxCharNumber: 256,
```
chamkank: My bad this should be 254 because we don't need the angle brackets https://stackoverflow.com/a/574698

PR #110, opened 2018-09-15
Add service worker

### LINKED ISSUE #105: Add service worker

chamkank:
:tickets: **Ticket(s)**: Closes #105 

---

Added a simple service worker.

## :thought_balloon: Notes

The service worker currently caches the root directory. Any fetch requests are intercepted to check if the resource is already in cache, and if not, it continues to fetch them from the network. 

All testing was done on localhost, honestly not sure how this will work in prod.

In the Chrome Dev tools, check to see that the service worker is registered. If the website content changes, a new service worker should be installed.

Service worker:
image

Cache:
image

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -15,7 +15,7 @@
   <link rel="apple-touch-icon" sizes="144x144" href="/assets/favicon/apple-icon-144x144.png">
   <link rel="apple-touch-icon" sizes="152x152" href="/assets/favicon/apple-icon-152x152.png">
   <link rel="apple-touch-icon" sizes="180x180" href="/assets/favicon/apple-icon-180x180.png">
-  <link rel="icon" type="image/png" sizes="192x192"  href="/assets/favicon/android-icon-192x192.png">
+  <link rel="icon" type="image/png" sizes="192x192" href="/assets/favicon/android-icon-192x192.png">
   <link rel="icon" type="image/png" sizes="32x32" href="/assets/favicon/favicon-32x32.png">
   <link rel="icon" type="image/png" sizes="96x96" href="/assets/favicon/favicon-96x96.png">
   <link rel="icon" type="image/png" sizes="16x16" href="/assets/favicon/favicon-16x16.png">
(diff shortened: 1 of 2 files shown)


web/index.html
```diff
@@ -29,6 +29,33 @@
 
 <body>
   <div id="app"></div>
+  <script>
+    if ('serviceWorker' in navigator) {
+      window.addEventListener('load', function () {
+        navigator.serviceWorker
+          .register('sw.js', { scope: '/' })
```
d4l3k: You probably want to change this to match the same name as the old one so it'll overwrite the existing one.

https://github.com/nwhacks/nwhacks2018_static/blob/master/index.html#L26

web/index.html
```diff
@@ -29,6 +29,33 @@
...
+          .then(function (registration) {
+            registration.onupdatefound = function () {
+              var newWorker = registration.installing;
+              newWorker.onstatechange = function () {
+                switch (newWorker.state) {
+                  case 'installed':
+                    if (navigator.serviceWorker.controller) {
+                      console.log('New website version available, reload to update.');
```
d4l3k: Should make this into a toast/snackbar instead so users will know to reload.

web/sw.js
```diff
@@ -0,0 +1,17 @@
+const CACHE_NAME = 'nw-cache';
+const CACHED_ROUTES = [
+  '/',
+];
+
+this.addEventListener('install', (e) => {
+  e.waitUntil(
+    caches.open(CACHE_NAME).then(cache => cache.addAll(CACHED_ROUTES))
```
d4l3k: This seems extremely aggressive and I'm not sure it'll ever invalidate the cache. Since sw.js is static I don't think the service worker will ever be updated.

chamkank: Ah you're right, seems that it stops on the installing stage. I think if I piped some dummy statement (`var a = random_string`) to the end of `sw.js` in our build process it would fix the issue. But using a plugin does seem a lot cleaner. :thinking:

chamkank:
Thanks for the feedback, I'm gonna try changing this to use the first plugin you suggested. :+1:

chamkank:
Closing this until our November milestone tasks are finished & I find time to implement this

PR #108, opened 2018-09-15
Web/home/FAQ

chamkank:
:tickets: **Ticket(s)**: Closes #74 

---

## :construction_worker: Changes

This PR implements the FAQ section.

## :thought_balloon: Notes
- The emojis in the headers are actually SVG files [to make sure that they look consistent on all devices](https://emojipedia.org/school-satchel/)
- The content is still missing for a few of the questions, these will either be filled in before the PR gets merged, or the incomplete questions will be removed entirely and we'll update add them back incrementally as we get the answers.

## :flashlight: Testing Instructions

Check to see that it looks right on different screen widths. I've tested it between the 320px - 2560px screen width range on Chrome and FireFox.

Issue #76, opened 2018-08-30
Target 1080p for standard homepage layout, fix padding


bobheadxi:

bobheadxi:
From #92 (@sherryyx ):

> I know there was a bug logged previously about the text and the illustrations not being side by side on macbook pros. Now there's another bug where when the illustration appears below the text, there's a margin-right on the text that's making it horizontally off-centered.

![image](https://user-images.githubusercontent.com/29152180/45009345-84b58080-afbc-11e8-83a2-52cb73582b80.png)

chamkank:
I can take this on once FAQ and footer are merged into master so we can fix it all at once

chamkank:
Also adding to this issue:
- Typo fix (600 -> 650)
- Consistent padding between sections

PR #116, opened 2018-09-25
Add FAQ to navbar

chamkank:

Small change, forgot to add it in #108


chamkank:
Yeah I noticed that earlier as well, I'll add extra padding to the bottom of stories, along with the fixing the other padding issues in #76

Issue #70, opened 2018-08-25
no-alert?


bobheadxi:
> JavaScript’s alert, confirm, and prompt functions are widely considered to be obtrusive as UI elements and should be replaced by a more appropriate custom UI implementation... This rule is aimed at catching debugging code that should be removed and popup UI elements that should be replaced with less obtrusive, custom UIs.

https://eslint.org/docs/rules/no-alert

Relevant code: https://github.com/nwhacks/nwhacks2019/blob/7436578f3a2f1177463cbd0ef80323a15df597f1/web/components/application/hacker/Hacker.js#L53

Thoughts, @mingyokim ?

mingyokim:
I’d say alerts are fine as long as design team is fine with it, since we don’t have the custom UI for alerts ready to use right now. Alerts would help us move forward faster to have an MVP without compromising functionality

sherryyx:
@bobheadxi Yeah let’s just leave it for
now!

PR #121, opened 2018-10-07
Web/cache/delete

### LINKED ISSUE #120: Remove old service worker
And eventually add a new one with #105

chamkank:
:tickets: **Ticket(s)**: Closes #120 

---

Updated service-worker.js to unregister the 2018 service worker and its cache. This will eventually be replaced with a new service worker through #105 but we'll go with this for now so that the website works (for people with the old service worker) when we launch it on Sunday.

This one is tricky to test locally unless you've already run the old website on localhost with the same port. 

So here's a demo from my machine:


chamkank:
That sounds like a better idea, I've made it so it forces the reload (see https://github.com/nwplus/nwhacks2018_static/pull/92).

chamkank:
I was a fan of the notification because it made sense for smaller updates, so that the user knows that something has changed. But since the site is completely new, the forced reload seems more convenient (assuming that the forced reload works on all devices 🤞 ).

However for future updates to the 2019 site, I think the notifications would be nice. I'll try to ship #110 at the same time as #70.

PR #123, opened 2018-10-07
Fix homepage layout & padding

### LINKED ISSUE #76: Target 1080p for standard homepage layout, fix padding

### LINKED ISSUE #118: Remove horizontal scrollbar on homepage
There's a horizontal scrollbar on the page - one or more elements are taking up more than 100% screen width.

chamkank:
:tickets: **Ticket(s)**: Closes #76, fixes #118 

---

1. Made the first & second panels look closer to the mockups.
2. Homepage targets 1080p width instead of 1440p.
3. Fixed the padding & other issues mentioned in #76.
4. Removed the horizontal scrollbar.
5. Comment out sign in button.
6. Added metatags (`og:image`, `og:description`, `og:url`).

Test on different screen widths.


web/styles/typography.sass
```diff
@@ -60,14 +60,14 @@ p
     font-size: 24px
     line-height: 36px
     margin-bottom: 40px
-    font-style: bold
+    font-weight: bold
```
mingyokim: not sure if this is the cause but the headers don't seem to be bolded anymore:

image

web/components/navbar/index.js
```diff
@@ -2,24 +2,24 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import { DISPLAY_TYPE } from '../../containers/navbar/DisplayTypes';
-import { BUTTON_TYPE } from '../../containers/navbar/ButtonTypes';
+// import { BUTTON_TYPE } from '../../containers/navbar/ButtonTypes';
```
mingyokim: Why are we commenting out these lines? It shouldn't show up on production because they're all hidden under the "application" feature flag anyway.

web/styles/typography.sass
chamkank: What browser is this? On my end it looks bold on both Chrome and Firefox

image

mingyokim: oh it was on safari LOL

web/components/navbar/index.js
mingyokim: yeah that's probably because the "application" feature flag was enabled in `nwhacks2019-dev` when you saw the sign in button

PR #129, opened 2018-10-09
Web/stories/fix breakpoint

mingyokim:

---

Support the hover effect on more viewports

## :thought_balloon: Notes


web/styles/breakpoints.sass
```diff
@@ -125,7 +131,10 @@ $mobile-s:  320px
     &-mobile
       @media screen and (min-width: $mobile)
         display: none !important
-      
+    &-stories
```
chamkank: Maybe we should put `scale-show-stories` and `scale-hide-stories` in `home/index.sass` because it seems like a really specific style that we won't reuse.

web/styles/breakpoints.sass
```diff
@@ -152,3 +161,15 @@ $mobile-s:  320px
...
               cursor: default !important
               opacity: 100 !important
+      &-desktop
+        &:hover
+          @media (hover:hover)
+            @media screen and (max-width: $desktop)
+              cursor: default !important
+              opacity: 100 !important
+      &-stories
+        &:hover
```
chamkank: Ooo nice, I think putting `&:hover` before the `@media (hover:hover)` avoids the Firefox hover bug.

PR #135, opened 2018-10-19
make a page template and don't include sign up page

mingyokim:

---

* Refactored pages by using a page template - the latter contains `ProgressGroup` as well as next/prev buttons, and these components are used in every page
* Fixed background colour being cut off (you can notice this at /application/hacker in the smallest mobile view) with some css magic??
* Removed `SignUp` page from the list of pages being supplied to `HackerApplication`
* Made abstract class for `Page`
* Tried to fix `NotFound` component (the mountain icon was cut off by the navbar so added navbar padding)

## :thought_balloon: Notes

* Test background colours of hacker application pages & other pages as well
* The hacker application pages functionality should be the same as before


web/components/application/hacker/pages/PageOne.js
```diff
@@ -1,102 +1,67 @@
...
+                As expected, we receive a large number of applications every year.
+                We focus on curating a quality hackathon experience for each attendee.
+                For a fair assessment of your application,
+                we encourage you to put your best foot forward on this journey!<span role="img" aria-label="mountain">⛰️</span>
+                </p>
+                <TextInput
+                  label="First Name"
+                  name="first-name"
+                  value={firstName}
+                  onChange={(fname) => {
```
chamkank: Correct me if I'm wrong, but when we add more components to this page, wouldn't a similar `onChange` function be required for every required input field in the page? We should probably make a more generic function that we pass to `onChange` that takes the field name and value as params.

web/components/application/hacker/pages/PageTemplate.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { ProgressGroup, SecondaryButton, PrimaryButton, ButtonGroup } from '../../../input/buttons';
+import { getPrimaryButtonText } from './utils';
+import { ApplicationContext } from '../../../../containers/application/Context';
+
+export const HackerApplicationPageTemplate = props => (
```
chamkank: 💯

web/components/application/hacker/pages/Page.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export class Page extends React.Component {
```
chamkank: Awesome abstraction 🙏 do you think we could move this to `web/components/application` so we don't have to duplicate it when doing the volunteer application?

web/components/application/hacker/pages/PageOne.js
mingyokim: good point!!

PR #139, opened 2018-10-20
make 404.html

mingyokim:

---

So I've noticed that our `page_not_found` route doesn't work on `nwhacks.io` (although it works locally). You can also duplicate this issue by running `yarn serve`.

This problem is well described here: https://github.com/rafrex/spa-github-pages. But in summary:

> Why it's necessary
> GitHub Pages doesn't natively support single page apps. When there is a fresh page load for a url like example.tld/foo, where /foo is a frontend route, the GitHub Pages server returns 404 because it knows nothing of /foo.

The linked repo above provides a solution, where you create a `404.html` that redirects to `index.html` with a really complicated script. You can read more about it in the repo, and I've verified that it works, but I wasn't too down because the scripts looked overly complicated.

So I thought about it for a sec and i was like why redirect if you can just render the application in `404.html`??? And it worked!!

## :thought_balloon: Notes

Try `yarn serve` and go to `page_not_found`. It should return our `NotFound` component instead of the default 404


web/webpack.plugins.js
```diff
@@ -11,6 +11,11 @@ module.exports = [
     filename: 'index.html',
     inject: true,
   }),
+  new HtmlWebpackPlugin({
```
chamkank: Is the change to `webpack.plugins.js` needed? I tried it without this change it and it still seems to work.

mingyokim: @chamkank Yes, we need it to make `404.html` in `docs/`.

If you delete `404.html`, run `yarn build`, and run `yarn serve`, it shouldn't work because `404.html` wouldn't be generated. Also if you change `index.html` then this would be needed to update `404.html` as well.

chamkank: > Also if you change index.html then this would be needed to update 404.html as well.

Ah I didn't think of that

PR #141, opened 2018-10-21
Web/input/drop down list

### LINKED ISSUE #112: Implement regular + searchable dropdown input component
From Figma:
image

chamkank:
:tickets: **Ticket(s)**: Closes #112 

---

This implements the regular & searchable dropdown list components using [react-select](https://github.com/JedWatson/react-select).

## :thought_balloon: Notes
If you're curious, here are the different ways that I found to implement this component and the problems with them: 

### Using `<select>`
Normally a dropdown is implemented with the `<select>` and `<option>` HTML elements like so:
```
<select ...>
  <option ...>A</option>
  <option ...>B</option>
  <option ...>C</option>
</select>
```
And to make it searchable, you could overlay an `<input type=text>` field on the dropdown and automatically filter the list with an `onChange` function. However the issue I ran into is the fact that `<select>` and `<option>` have some permanent default styles that are implemented by the browser. One of these styles are `option:hover` which [cannot be overridden in Chrome](https://codepen.io/xram/pen/xyoLn). i.e. there's no way to get the custom option hover color seen in the mocks with this approach. 

### Using `<datalist>`
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist
This would have been nearly perfect - it suffers from the same issue mentioned above BUT it has all the search logic already written (i.e. supposed to be implemented by the browser). Except it's not supported in Safari at all. 😢 

### Creating a custom construct
The last option was to either use `<ul><li>` or `<div>` to manually implement a component that looks and behaves like a typical select component. This would involve writing all of the logic that the browser already implements for the `<select>` component (the dropdown menu, it's animation, etc) plus the extra search logic we need to add. The most complicated solution but it doesn't have the problems found in the other approaches.

### What I actually used
Turns out that there are 20 different variations that people have made using custom elements & logic: https://www.learningjquery.com/2017/09/20-impressive-jquery-plugins-to-replace-the-html-select-box

In particular I found [`react-select`](https://github.com/JedWatson/react-select/tree/master/src/__tests__) which is actively maintained and [well-tested](https://github.com/JedWatson/react-select/tree/master/src/__tests__). Here's what the final HTML structure looks like:

```
<div class="react-select-container">
  <div class="react-select__control">
    <div class="react-select__value-container">...</div>
    <div class="react-select__indicators">...</div>
  </div>
  <div class="react-select__menu">
    <div class="react-select__menu-list">
      <div class="react-select__option">...</div>
    </div>
  </div>
</div>
```
There is a hidden `<input>` element in the structure that stores the value so it works fine in a form.

The `<Select>` class I've created is a wrapper around the Creatable class from react-select, and styled to try and match the mocks.

Check out `/ui_demo` for 3 different versions of this component. For your convenience here's the code that renders the three `<Select>` components seen in the demo gif:

```
<Select
  name="some_select"
  options={['University of Toronto', 'McGill University', 'University of Waterloo', 'University of Alberta', 'University of Manitoba', 'University of Saskatchewan', 'University of New Brunswick', 'Memorial University of Newfoundland', 'Dalhousie University']}
  label="Regular drop-down input"
/>

<Select
  name="some_select"
  options={['University of Toronto', 'McGill University', 'University of Waterloo', 'University of Alberta', 'University of Manitoba', 'University of Saskatchewan', 'University of New Brunswick', 'Memorial University of Newfoundland', 'Dalhousie University']}
  label="Searchable drop-down input"
  placeholder="Start typing to search"
  isSearchable
/>

<Select
  name="some_select"
  options={['University of Toronto', 'McGill University', 'University of Waterloo', 'University of Alberta', 'University of Manitoba', 'University of Saskatchewan', 'University of New Brunswick', 'Memorial University of Newfoundland', 'Dalhousie University']}
  label="Searchable drop-down input that allows new options"
  formatNewOptionLabel={label => `My school is not listed, use "${label}"`}
  placeholder="Start typing to search"
  isSearchable
  allowNewOption
/>
```


chamkank:
Yup exactly, you can pass in something like 
`onChange={(props) => { console.log(props); }}` 
and when a new option is selected it will spit out
`{label: "some label", value: "it's value"}` 

If a user-defined option is selected as the value it will include a `__isNew__` field
`{label: "some label", value: "it's value" {label: "ddsadsadasdasdas", value: "ddsadsadasdasdas", __isNew__: true}}`

PR #142, opened 2018-10-24
add validations to each application page

mingyokim:

---

I quickly realized that the following code does not scale to the number of fields we have on the page.

```
// This is fine when we just have one field
shouldNextButtonBeEnabled = () => {
    const {	   
      application: {	   
        firstName,	
      },	
    } = this.props;	
    return firstName.length > 0;	
  }
```

For example, if we had many more fields, we would have to validate each field manually

```
// Starts getting annoying when we start adding a lot more fields 😩 
shouldNextButtonBeEnabled = () => {
    const {	   
      application: {	   
        firstName,
	lastName,
        email,
        ....,
      },	
    } = this.props;	
    return firstName.length > 0 && lastName.length > 0 && email.length > 0 && ....;	
  }
```
So I changed it so that all you have to do is the following:

```
constructor(props) {
    super(props);
    this.constraints = constraints;
}
```
Where `constraints` is a simple js object defining the validations. (https://validatejs.org/#validators)

e.g.

```
{
   firstName: {
      presence: true,
      length: {
          minimum: 1,
          message: 'cannot be empty.',
      },
   }
}

```

Furthermore, you have to pass in additional props to the text input:
```
onBlur={() => this.setFieldAsBlurred('firstName')} // keep track of which fields are blurred - after they are blurred, display the error below if any
error={this.getErrorIfBlurred('firstName')} // pass in the error related to this field
```

## :thought_balloon: Notes

- Make sure the next button is enabled only when all fields are validated
- Make sure the error is only displayed when you "blur"/click out of the input
- Make sure the error goes away when you fix it
- Test behaviour when refreshing the page


web/yarn.lock
```diff
@@ -7927,6 +7927,10 @@ validate-npm-package-license@^3.0.1:
     spdx-correct "^3.0.0"
     spdx-expression-parse "^3.0.0"
 
+validate.js@^0.12.0:
```
chamkank: This dependency gets locked when I run `make`, not sure why it's different here

PR #143, opened 2018-10-25
Web/improve/select

chamkank:
:tickets: **Ticket(s)**: Improves #112 

---

This exposes the following event handler arguments to react-select:
- onBlur
- onFocus
- onInputChange
- onKeyDown
- onMenuOpen
- onMenuClose
- onMenuScrollToTop
- onMenuScrollToBottom,

With this change I've added an example to the ui_demo that shows `onInputChange` in use to replace the options with random options every time the event is fired.

Take a look at `/ui_demo` but other than that this is already tested behavior in the react-select library.


chamkank:
Commits will be merged with #144

PR #150, opened 2018-10-28
Web/application/hacker/pages/page two

StanfordLin:

---

-Added most of the features for PageTwo
-Modified the operator under web/components/input/text/TextArea so that if there are 250 words, it will appear as "You have 0 characters left" in blue.
-Hacked the behaviour of TextArea so it counts words. 
```
     onTextChange = (text) => {
                        this.setState({
                          currentCharCount: text.match(/(\w+)/g).length,
                        });
                      };
```

## :thought_balloon: Notes

Need help on:

1. Making resume link mandatory, and adding the question hover help
2. Determining checkbox data structure for the roles ppl are interested in and how we want to store it's values

3. Fixing WordCount. Builtin TextArea only has character count. Not sure if you guys want to create a new custom component to count words or to modify TextArea to support word count.

I created a duplicate component of TextArea under  web/components/input/text/TextAreaWordCount but I kept running into export/undefined error. Attached is the screenshot.
screen shot 2018-10-27 at 9 36 50 pm

4. Need to change the display so that it says "You have X words left instead of characters.", but going to wait because it kinda depends on how we fix word count.

5. Do I need to convert the emoji's to svgs? How would I add svgs within quotes?

All these todos can be found by searching TODO in web/components/application/hacker/pages/PageTwo/PageTwo.js

Go to http://localhost:8080/#/application/hacker and click the number 2


chamkank:
Just previewed it, it looks nice!

> 1. Making resume link mandatory, and adding the question hover help

For making fields required, @mingyokim set up validate.js, You would define a constraints file (ex: [web/components/application/hacker/pages/PageOne/PageOneConstraints.js](https://github.com/nwplus/nwhacks2019/blob/master/web/components/application/hacker/pages/PageOne/PageOneConstraints.js)), import it to `PageTwo.js` and add it to the constructor (see [PageOne.js for an example](https://github.com/nwplus/nwhacks2019/blob/master/web/components/application/hacker/pages/PageOne/PageOne.js#L19)).

As for the hover help, we've backlogged it as it's not super important, if you're keen on implementing this you can use export the question mark svg from Figma and use a third-party tooltip library (ex: [tippyjs](https://github.com/atomiks/tippyjs)) to reveal a tooltip on hover. If you don't have time for this, you can change the label for the resume link from `Resume` to `Resume Link (make sure the link is publicly accessible)`.

> 2. Determining checkbox data structure for the roles ppl are interested in and how we want to store it's values

> 3. Fixing WordCount. Builtin TextArea only has character count. Not sure if you guys want to create a new custom component to count words or to modify TextArea to support word count.
> 4. Need to change the display so that it says "You have X words left instead of characters.", but going to wait because it kinda depends on how we fix word count.

Oh for this we decided earlier for `TextArea` that we would be using max characters instead of word count because it's hard to regulate the length of words. So you can still use TextArea, but you should ask logistics what the character limit should be for each of the long answer fields.

StanfordLin:
Cool, made some more changes.

1. How did you find/generate an svg of an emoji?
2. When I tried using constraints on TextArea, it doesn't appear to be updating on change
3. Not sure how to implement the checkbox group? Also, there is no label in the checkbox group even though a prop of label is passed through it

chamkank:
> 1. How did you find/generate an svg of an emoji?

What I did for the FAQ emojis in Figma was: 
- Copy the emoji from some textbox. 
- Create a new textbox (the icon is on the toolbar) and paste the emoji.
- Click out of the textbox then click on it again and export it as an svg.

I'll answer 2 and 3 in the morning. 💤

web/components/application/hacker/pages/PageTwo/PageTwo.js
```diff
@@ -17,36 +19,159 @@ class PageTwo extends Page {
...
+          <Checkbox
+            label="Other"
+            value="true"
+            checked={isOther === true}
+          />
+
+        </CheckboxGroup>
+
+        <TextInput
+          label="Github/BitBucket/Github link (optional)"
```
chamkank: `Github/BitBucket/Github link` --> `Github/BitBucket/Gitlab link`

web/components/application/hacker/pages/PageTwo/PageTwo.js
```diff
@@ -17,36 +19,159 @@ class PageTwo extends Page {
...
+            this.updateApplication({ isHacker: isCheckboxChecked });
+            this.updateApplication({ isDesigner: isCheckboxChecked });
+            this.updateApplication({ isHardware: isCheckboxChecked });
+            this.updateApplication({ isOther: isCheckboxChecked });
+          }
+          }
+        >
+          <Checkbox
+            label="Developer"
+            value="true"
```
mingyokim: The value for all your checkboxes are the same. You need to make them unique per option.

One way to do this:

1. Create an object that maps the Checkbox value to a key in the application
```
const map = {
  hacker: "isHacker",
  designer: "isDesigner",
}
```

2. In each `Checkbox`, pass in the appropriate key from the map in the "value" prop
```
<Checkbox
  label="Developer"
  value="hacker"
  ...
```

3. In `onChange` callback function of `CheckboxGroup`, the argument you are getting is the "value" of the selected `Checkbox`. So all you have to do is:
`this.updateApplication({ map[selectedValue]: true })`

This doesn't handle "unselecting" a checkbox though, so you'd have to figure that out yourself for now. I'm not exactly sure what argument is supplied in the `onChange` function, maybe @chamkank has a better a better idea. Or you can also use the chrome debugger tool, and set a breakpoint in this callback function. If you don't know how to use the chrome debugger tool, just ping me and we can set up a call

chamkank: You could also give each checkbox its own `onChange` function instead of passing a single one down through the CheckboxGroup. This is probably the easiest approach. 

For example, the Developer checkbox would be:
```
<Checkbox
  label="Developer"
  value="true"
  className="margin-left-none"
  onChange={(e) => {
    this.updateApplication({ isDeveloper: e.target.checked });
  }}
/>
```

web/components/application/hacker/pages/PageTwo/PageTwo.js
```diff
@@ -17,33 +20,161 @@ class PageTwo extends Page {
...
+          value={interestForNwHacks}
+          maxLength={750}
+          rows={7}
+          onChange={newInterestForNwHacks => this.updateApplication({
+            interestForNwHacks: newInterestForNwHacks })}
+          onBlur={() => this.setFieldAsBlurred('interestForNwHacks')}
+          error={this.getErrorIfBlurred('interestForNwHacks')}
+        />
+        <TextArea
+          label={(<div><img className="vertical-align-top emoji" alt="🏅" src={medal} /> Tell us about a recent project you&apos;ve worked on that you&apos;re proud of! (It doesn&apos;t have to be technical)</div>)}
```
mingyokim: for future reference, you can also do
`{Tell us about a recent project you've worked on that you're proud of! (It doesn't have to be technical)}`

mingyokim: then you don't need to replace all the ' with &apos;

PR #157, opened 2018-10-29
Web/function/submit application

chamkank:
:tickets: **Ticket(s)**: Closes #106 

---

## :construction_worker: Changes
- Added submission function to front-end
- Added submission function to back-end
- Added `make dev` and `make prod`
  - `make dev` will run `make web` and `make functions` in parallel, so it will locally host the front-end as usual and also emulate the cloud functions locally. All requests to cloud functions made from the front-end will automatically be made to the local cloud functions (which run on `localhost:5000`).
- `make prod` will run `make build` and `make deploy-prod` in parallel, so it will build the app into `/docs/` like usual, but also publish the cloud functions to our production environment

[Demo gif of `make dev` and submitting an application](https://i.imgur.com/eQKc3hD.gif)

## :thought_balloon: Notes (updated)
- reCAPTCHA is now fully implemented
- The function writes everything to two separate collections, `hacker_quick_info` and `hacker_main_info`. It also uses `validate.js`, and the constraints will need to be finalized with #158 

## :flashlight: Testing Instructions
- **[Update your env file, it now has our dev reCAPTCHA key](https://www.notion.so/nwplus/env-0407d7c9fc0849088d738c109ce654fc)**
- No idea how any of this will behave on a Mac, so try running `make dev`. If you get an error, check to see if the error is not already in `functions/README.md`. 
- Try submitting the application, and check your console for a success message, and also check firestore for new documents in the two collections

PR #171, opened 2018-10-31
Use the same ID for a given hacker's documents

chamkank:
:tickets: **Ticket(s)**: Fixes https://github.com/nwplus/nwhacks2019/pull/157#issuecomment-434510693

---

This makes a hacker's document have the same document ID in Firestore so we can associate the two.

When you submit the application, check the the ID of the document created in `hacker_quick_info` matches the document ID of the document created in `hacker_main_info`.


functions/src/submitApplication/index.js
```diff
@@ -36,12 +36,13 @@ exports.submitApplicationHacker = functions.https.onRequest((request, response)
       // create a batch job
       const batch = db.batch();
 
-      // create a new document in hacker_quick_info and set its value
+      // create a new document in hacker_quick_info and hacker_main_info with the same ID
       const hackerQuickInfoRef = db.collection('hacker_quick_info').doc();
-      batch.set(hackerQuickInfoRef, hacker.hacker_quick_info);
+      const hackerId = hackerQuickInfoRef.id; // use same document ID for this hacker's documents
```
mingyokim: What do you think about storing the id inside the doc as well? Might make things easier on the client js side

chamkank: That sounds like a good idea, I looked it up and [querying by document id seems kinda verbose](https://stackoverflow.com/a/48467056).

chamkank: On a similar note, should we store `hacker_quick_info` in `hacker_main_info` as well (we could call them `hacker_quick_info` and `hacker_full_info`)? i.e. `hacker_quick_info` would be sort of like a cache / index structure. 

The reason I mention this is because it would also make filtering easier. Ex: in our current set up filtering by `score > 8` and `travel_reimbursement == 2` would require us to check both documents since `score` would be in `hacker_quick_info`.

mingyokim: @chamkank yeah I agree, `hacker_full_info` should contain everything, and `hacker_quick_info` would only contain a subset of it

mingyokim:
@chamkank I'm getting this error? 
image

chamkank:
Oops yeah it is, let me merge that one and update this branch

functions/src/submitApplication/models.js
```diff
@@ -3,12 +3,14 @@ const cleaner = require('deep-cleaner');
...
   const hacker = {
     hacker_quick_info: {
+      id: data.id,
       firstName: data.firstName,
       lastName: data.lastName,
       email: data.email,
       timestamp: data.timestamp,
     },
     hacker_main_info: {
+      id: data.id,
```
mingyokim: Can we also add `hacker_quick_info` attributes to `hacker_main_info`? and maybe rename `hacker_main_info` to `hacker_full_info`?

Issue #159, opened 2018-10-29
Add loading animation, error, and success redirect to hacker application


chamkank:
handle loading/error/success when submit button is pressed

mingyokim:
closed by #157, #167, and #175

PR #175, opened 2018-11-01
Add loading animation and error handling

chamkank:
:tickets: **Ticket(s)**: Closes rest of #159 

---

After recaptcha token is generated and our cloud function is called, the submit button now becomes a "Submitting..." button. Also if our function fails for some reason, it will show an error message on the UI and put the submit button back into it's normal state.

+ also added unrelated change 277d2c7450b9bb191e7e5a863296d5006c765108

## :thought_balloon: Notes

This doesn't handle failing to contact the reCAPTCHA servers (i.e. if they're down or in the more likely case that the user is disconnected from the internet). In this case, the button never goes into the "Submitting..." state.

**[Here's a demo gif where I follow the below steps.](https://i.imgur.com/cpb44bQ.gif)**
1. Run `make web` and try to submit the application - it will fail because the submit function isn't being emulated (i.e. fails to response). You should see the "Submitting..." message followed by the error message on the UI. Try clicking on it multiple times to see that it still fails.
2. Run `make functions` in a separate terminal and now try the submit button. It should succeed.


web/components/application/hacker/pages/PageTemplate.js
```diff
@@ -46,15 +70,18 @@ export class HackerApplicationPageTemplate extends React.Component {
...
+              onClick={activeIndex !== count - 1 ? onPageNext : () => this.onSubmitButtonClick()}
             />
+            {submitFailed ? (<p>Failed to submit application, please try again!</p>) : null}
             <Reaptcha
               ref={e => (this.captcha = e)}
               sitekey={recaptchaConfig[process.env.NODE_ENV].sitekey}
               onVerify={(token) => {
-                submitApplication(null, token);
+                this.setState({ isSubmitting: true });
+                this.setState({ token });
```
mingyokim: Why is this necessary?

web/components/application/hacker/pages/PageTemplate.js
```diff
@@ -46,15 +70,18 @@ export class HackerApplicationPageTemplate extends React.Component {
...
+              text={isSubmitting ? 'Submitting...' : getPrimaryButtonText(activeIndex, count)}
+              onClick={activeIndex !== count - 1 ? onPageNext : () => this.onSubmitButtonClick()}
             />
+            {submitFailed ? (<p>Failed to submit application, please try again!</p>) : null}
             <Reaptcha
               ref={e => (this.captcha = e)}
               sitekey={recaptchaConfig[process.env.NODE_ENV].sitekey}
               onVerify={(token) => {
-                submitApplication(null, token);
+                this.setState({ isSubmitting: true });
```
mingyokim: @lynnzhang7 pointed out the delay, and I think that's because we're setting `isSubmitting: true` **after** recaptcha is verified. What do you think about doing this **before** recaptcha is verified, e.g. in https://github.com/nwplus/nwhacks2019/pull/175/files#diff-71f8e35c7373caaf57f69d2b381b566aR29?

web/containers/application/hacker/Hacker.js
```diff
@@ -31,7 +31,7 @@ export class HackerApplicationContainer extends React.Component {
     this.state = { isSubmitted: false };
   }
 
-  submitApplication = (userCredentials, recaptchaResponse) => {
+  submitApplication = (userCredentials, recaptchaResponse, onSubmitFailCallback) => {
```
mingyokim: love the use of callbacks here! 💪

web/components/application/hacker/pages/PageTemplate.js
chamkank: Ah that's what I initially did, the issue with this is that the case when the puzzle is triggered. If the puzzle was triggered then it would show `Submitting...` while the user is still doing the puzzle (which they can also click out of, leaving the submit button in the `Submitting...` state).

web/components/application/hacker/pages/PageTemplate.js
chamkank: If we're in this function it means that the token was successfully generated but the call to our cloud function may still fail. In this case, the recaptcha refuses to generate another token so I use the same token in L22 when the user tries to resubmit.

PR #176, opened 2018-11-01
Add confirm email field

chamkank:

---

Adds confirm email field with validation

You can try out the email and confirm email fields


web/components/application/hacker/pages/PageOne/PageOneConstraints.js
```diff
@@ -40,6 +40,13 @@ export const constraints = {
   email: {
     email: true,
   },
+  confirmEmail: {
```
mingyokim: I think we might need to add `presence: true` here, the next button is enabled for me when I don't fill in `confirmEmail` field

PR #183, opened 2018-11-05
Add 'non-binary' as gender option

xofeoz:

---

Just added 'non-binary' as gender option.

## :thought_balloon: Notes
(I know it could've been typed in before but it should be a 'default' option so to speak)


mingyokim:
I'm a little hesitant about adding a new option after we've opened application registration due to data integrity. Thoughts? @chamkank @StanfordLin @lynnzhang7

chamkank:
Hmmm I just checked out applicant data, as of right now we don't have any custom responses for gender (just `Male`, `Female`, and `Prefer not to specify`) so I don't think data integrity is a concern.

xofeoz:
I've had more than one friend who's part of the lgbtq+ community who's reached out to me to tell me how it may be interpreted by others - and I agree with them because I think being an inclusive and welcoming space is a bit more important than our data integrity since applications have been open for less than a week at this point

chamkank:
@xofeoz I've just invited to our nwplus org, in the future you can clone the repository and directly publish new branches to this repo (instead of opening pull requests from branches on xofeoz/nwhacks2019).

PR #192, opened 2018-11-22
Admin/auth

### LINKED ISSUE #190: Admin authentication
admin page authentication. Must includes:

* UI for sign in
* Correct authentication whenever visiting an admin page - if not logged in or not admin, redirect to 404

mingyokim:
:tickets: **Ticket(s)**: Closes #190 

---

* Restructure of the routes. Navbar only shows up for non admin pages, and all admin pages are wrapped by `AdminGate` component except for `admin/login` and `admin/logout`.
* Refactor `Login` component into a container and component 

## :thought_balloon: Notes

* Test that all the routes are still working
* Test the application is still working
* Test logging in (you can create a login on the firebase console)
* Test the admin routes - should not be able to access any admin routes other than admin/login and admin/logout (should redirect to page not found) if you're not logged in or if you're logged in but you're not an admin.


web/components/app/App.js
```diff
@@ -27,21 +28,51 @@ class App extends React.Component {
...
+          <Route path="/admin">
+            <Switch>
+              <Route path="/admin/login">
+                <Login />
+              </Route>
+              <Route path="/admin/logout">
+                <Logout />
+              </Route>
+              <Route path="*">
+                <AdminGate>
```
chamkank: This AdminGate thing is so clean 👍

web/components/auth/Login/Login.js
```diff
@@ -1,140 +1,69 @@
...
+              name="password"
+              showErrorMessage
+              showForgot
+              />
+            <PrimaryButton
+              text="Submit"
+              type="submit"
+              className="fill-width"
+              />
+            <p className="flex jc-center">
```
chamkank: Can we comment this out for now to avoid any confusion?
https://github.com/nwplus/nwhacks2019/blob/2bf35852b128dcdd2d95f60d0d970f79ad3803e7/web/components/auth/Login/Login.js#L48-L51

web/components/admin/AdminGate/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const {
+    data: {
+      admins,
+    },
+  } = firestore;
+
+  const { uid } = auth;
+  const isAdmin = uid in admins;
```
chamkank: `uid in admins` is throwing an error when `uid` is `undefined` and `admins` is `null` (i.e. when you're not logged in and try accessing any of the admin routes).

Specifically: `Uncaught TypeError: Cannot use 'in' operator to search for 'undefined' in null`

web/components/admin/AdminGate/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+  if (isEmpty(auth)) {
+    return (<Redirect to="/page_not_found" />);
```
chamkank: Maybe we should redirect to `/admin/login` here instead of `page_not_found`

web/components/admin/AdminGate/index.js
mingyokim: o good catch

web/components/admin/AdminGate/index.js
mingyokim: hm maybe, not convinced why /admin/login is strictly better than page_not_found

web/components/admin/AdminGate/index.js
mingyokim: we don't need to check whether `uid` is undefined/null though, because `null in {...}` is false

web/components/admin/AdminGate/index.js
chamkank: This condition is triggered when the user is trying to access an admin page, we can assume that it's an organizer who's trying to access a page but has forgotten to log in. In this case it would save them the trouble of navigating to `/admin/login` manually.

web/components/admin/AdminGate/index.js
chamkank: Oh true

web/components/admin/AdminGate/index.js
chamkank: Ex: navigating directly to `nwhacks.io/#/admin/assessment` through the dropdown history on a URL bar

mingyokim: Yeah I agree, but it also makes the `admin` pages more likely to be noticed by non admin users. The behaviour of non admin users are unpredictable, but the behaviour of < 30 admin users is (i.e. we can explicitly say they need to login first at `/admin/login` before having access to the admin features)

chamkank: True but I don't see an issue with them discovering `/admin/` especially since if they were really trying to snoop around they would probably be looking at this repo for the routes. Redirecting to login seems like it would improve the UX experience for organizers instead of forcing them to use `/admin/login` each time or making them double check that they have no typos in their URL.

mingyokim: Well, it's not just about malicious users, it's also about the experience of the non admin users who accidentally stumble upon the admin section and they are prompted to login, at which point they would be very confused and uncomfortable.

I think we can both agree that this is just a UX decision, and I don't have enough expertise to make the call on which user experience is more valuable (organizers vs non-admin users). What do you think about consulting the design team?

chamkank: Yeah that sounds like a good idea

Ciuffi: I'm a little late but some thoughts: If a normal user "accidentally" finds the login page then its either a crazy coincidence or we don't have url routes that make sense. I think its better UX to expect someone visiting a url to want to be there (which hangs on the fact that we have good routes/good UX in other parts of the website). A simple "back to nwhacks" would also help reduce confusion. I also don't think this is worth changing unless it causes problems in the future though.. just wanted to add my 2 cents.

Issue #194, opened 2018-11-23
Backfill quickinfo


mingyokim:
Add all the fields except for the long responses to the hacker_quick_info collection, for all the applications submitted before we made the change in the Cloud function.

PR #195, opened 2018-11-23
add more fields to quick info

mingyokim:

---

* Fixed bug where we were not submitting `birthday` and `isPrivacyPolicyChecked`
* Two new collections, `hacker_short_info` and `hacker_long_info`.
* `hacker_short_info` contains all the fields + id and timestamp except for long responses.
* `hacker_long_info` contains the long responses + id and timestamp. 
* `hacker_full_info` is the same as before (but with bug fix) - contains all the fields + id and timestamp

## :thought_balloon: Notes

This only applies to the applications submitted after we merge this in. So we need to backfill (#194)

- [ ] Test that full info contains all the required fields (e.g. birthday!!) + id and timestamp
- [ ] Test that short info contains all the required fields + id and timestamp, except for long responses
- [ ] Test that long info contains all the long responses + id and timestamp
- [ ] Test that all documents created in full, short, and long info all have the same ID.


chamkank:
Can we rename `hacker_quick_info` to `hacker_short_info` or `hacker_main_info`? I think it would be more appropriate. And can we store `interestForNwHacks` and `recentProject` in a new `hacker_long_info` collection? We can leave `hacker_full_info` as a backup just in case using two collections somehow goes wrong.

mingyokim:
@chamkank ready for review

functions/src/submitApplication/constraints.js
```diff
@@ -115,4 +115,168 @@ exports.constraints = {
     },
     recaptchaResponse: atLeastOneCharacter,
   },
+  hacker_full_info: {
```
chamkank: If we're going to be validating the constraints on a per-collection basis instead of the form data as a whole, can we remove the generic `hacker` constraints?
https://github.com/nwplus/nwhacks2019/blob/47436eaa73bc5c5f8ee9007cf0a6d64b8209dbbb/functions/src/submitApplication/constraints.js#L25-L117

functions/src/submitApplication/models.js
```diff
@@ -13,7 +13,7 @@ exports.Hacker = function Hacker(data) {
       city: data.city,
       school: data.school,
       gender: data.gender,
-      birthDate: data.birthDate,
+      birthDate: data.birthday,
```
chamkank: Can we rename `birthDate` to `birthdate`? As well as replace every `birthday` in the project (in web and functions) with `birthdate`? I intended to do that when I initially created `model.js` but I messed up by not updating `birthday` in the front-end, and by capitalizing the D in `birthdate` 😕

web/components/application/hacker/pages/PageTemplate.js
```diff
@@ -19,19 +19,12 @@ export class HackerApplicationPageTemplate extends React.Component {
...
-    const { submitApplication } = this.props;
-    if (submitFailed) { // if user has previously failed a submit
-      this.setState({ isSubmitting: true, submitFailed: false });
-      const { token } = this.state; // still-valid token from the last time user clicked submit
-      submitApplication(null, token, this.onSubmitFail);
-    } else {
-      this.setState({ submitFailed: false });
-      this.captcha.execute();
-    }
+    this.setState({ isSubmitting: true, submitFailed: false });
```
chamkank: Lol thanks for doing this, looks a lot nicer than before

functions/src/submitApplication/constraints.js
mingyokim: No, we need all these constraints because we need to be able to distinguish between whether the input to the cloud function is valid (if invalid, error code is 400, meaning the user didn't pass in the correct data) vs whether the transformed data to store in firestore is valid (if invalid, error code is 500, meaning the server's fault).

functions/src/submitApplication/models.js
mingyokim: I think that's getting out of the scope of this PR (because this PR is just focused on changing the way we store the hacker application in firestore) - we could make a separate issue for it. I personally think that's a low priority task bc it doesn't add any value/functionality for end users, and especially given that we have a very tight deadline. What do you think?

functions/src/submitApplication/constraints.js
chamkank: Oh got it, that makes sense

functions/src/submitApplication/models.js
chamkank: I think we should fix at least the capitalization of `birthDate` so we don't have `birthdate` and `birthDate` in our database if we end up doing the task.

mingyokim: Why would we have both `birthdate` and `birthDate` in our database? I'm not changing the name of the field in firestore, it's still `birthDate` like before

chamkank: Yeah I'm saying that we should rename it from `birthDate` to `birthdate` lol

mingyokim: Oh.. but why do you want to change it? I’m just reluctant bc we would have to backfill. If we change it from birthDate to birthdate then all the applications before this PR have the birthDate field while all the applications after this PR will have the birthdate field instead?

chamkank: The `birthDate` field wasn't written to the DB at all because of the typo right? Previously it was       `birthDate: data.birthDate` but `data.birthDate` was actually `data.birthday` on the front-end, so the field ended up being null and removed by deep-cleaner before the database write.

chamkank: So I see this as a chance to fix the grammar before we start writing the field, super trivial but yeah

mingyokim: Ohhhhh lol yeah I totally forgot, forget what I said about backfilling

PR #196, opened 2018-11-24
Web/internal/assessment panel

### LINKED ISSUE #191: Initialize admin dashboard layout & display list of hackers
- Empty three panel layout & toolbar for applicant assessment admin page
- Ability to view list of hackers

chamkank:
:tickets: **Ticket(s)**: Closes #191 

---

- Initialized the applicant assessment page with three-column layout and some styling
- Implemented ability to see list of applicants
- Implemented ability to see full info about an applicant (in plain text)

[Initial demo gif](https://i.imgur.com/pAUHagT.gif)

### After revisions:
- There are now buttons to switch between applicant type
- `ApplicantInfo` takes in applicantId and applicantType and fetches `shortInfo` and `longInfo` (if not already in the store)
- ScorePanel is not implemented but exists and takes in `applicantId` and `applicantType` and fetches `shortInfo` (if not already in the store)
- Implemented dropdown to switch between applicant types in Toolbar

## :thought_balloon: Notes

Overview:
-  **Container:** `web/containers/admin/AssessmentPage/index.js`: Handles logic for assessment page
    - Calls **component** `web/components/admin/AssessmentPage/index.js` which handles assessment page rendering
      - **Toolbar** (`web/components/admin/Toolbar/index.js`)
        - created component, but not implemented
      - **Applicant List**: (`web/components/admin/ApplicantList/index.js`) ✓
      - **Score panel** (`web/components/admin/ScorePanel/index.js`)
        - created component, but not implemented
      - **Applicant Info**: (`web/components/admin/ApplicantInfo/index.js`)
        - created component, not implemented (shows unformatted info for now)

Important notes:
- ~I'm using `hacker_full_info` in place of `hacker_main_Info` (until that's created)~
- Long answers are fetched from `hacker_long_info`

1. Login through `admin/login` using the [admin test credentials](https://www.notion.so/nwplus/Secrets-Configuration-ab7a6c0e31a24803a923759d16068dce#8d72de2ecfb443949cd8c82b7d1f1b84)
2. Navigate to `admin/assessment`
3. Click on some applicants.

The `AssessmentPageContainer` works for all types of applicants (`hackers`, `mentors`, `volunteers`). ~I haven't implemented the container method that switches between these collections but if you still want to test it, you can change the following value to `mentor`~


web/components/admin/ApplicantList/index.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/* eslint-disable camelcase */
+
+class ApplicantList extends React.Component {
+  render() {
+    const { applicants, selectedApplicantId, onApplicantClick, className } = this.props;
+    let applicantNumber = 0;
```
mingyokim: hm not sure about the purpose of `applicantNumber`. Would it be more helpful to display the id instead?

web/components/admin/ApplicantList/index.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/* eslint-disable camelcase */
+
+class ApplicantList extends React.Component {
```
mingyokim: Nice! 💪

web/containers/admin/AssessmentPanel/index.js
```diff
@@ -0,0 +1,131 @@
...
+    {
+      collection: 'mentor_full_info',
+      storeAs: 'mentorMainInfo',
+      orderBy: 'timestamp',
+    },
+  ]),
+  connect((state) => {
+    return {
+      firestore: state.firestore,
+      // auth: state.auth,
```
mingyokim: remove this?

web/containers/admin/AssessmentPanel/index.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+AssessmentPanelContainer.propTypes = {
+  firestore: PropTypes.object,
+};
+
+export default compose(
+  firebaseConnect(),
+  firestoreConnect([
```
mingyokim: Why are we fetching all the collections? Can't we just fetch the hacker collection first, and then fetch the rest if the user asks for it?

web/containers/admin/AssessmentPanel/index.js
```diff
@@ -0,0 +1,131 @@
...
+      applicants: {},
+      // single document from <applicantType>_main_info for selected applicant
+      selectedApplicantMainInfo: {},
+      // single document from <applicantType>_long_info collection for selected applicant
+      selectedApplicantLongInfo: {},
+    };
+  }
+
+  componentWillMount() {
+    this.setState({ applicants: this.getApplicants() });
```
mingyokim: should we also get the long info of the first applicant here by default?

web/components/admin/ApplicantList/index.js
```diff
@@ -0,0 +1,47 @@
...
+  render() {
+    const { applicants, selectedApplicantId, onApplicantClick, className } = this.props;
+    let applicantNumber = 0;
+
+    return (
+      <div className={`flex dir-col wrap ${className}`}>
```
mingyokim: image
Right now it's wrapping around into multiple columns, can we make it infinite scroll vertically?

web/containers/admin/AssessmentPanel/index.js
```diff
@@ -0,0 +1,131 @@
...
+    const { store } = this.context;
+    const { firestore } = store;
+
+    const longInfoCollections = {
+      hacker: 'hacker_long_info',
+      volunteer: 'volunteer_long_info',
+      mentor: 'mentor_long_info',
+    };
+
+    return firestore.get(
```
mingyokim: I've been noticing delays between clicking on applicants and displaying the info. 

I'm suspecting that `firestore.get` makes a new query every time without checking the cache, but let me know if I'm wrong.

One way to get around it is checking `this.props.firestore.data.hacker_long_info.[id]` and checking if we've already queried the id before. Then we don't need to call get.

But I think a more reliable/preferred way is to query "automatically" in whatever component is loading the long info. Then it would "automatically" load it using the best approach (by checking the cache/setting listeners, etc).

```
compose(
  firestoreConnect(({ props }) => {
    // get collection name from props, hacker/volunteer
    // get id from props
    return { collection: <collection name>, doc: <"id"> }
  }),
```

web/components/admin/ApplicantList/index.js
chamkank: We used the applicant number last year as a low-effort way to assign applicants to organizers (ex: "`________, can you mark resumes for applicants 100-200`"). 

If it helps we can show the applicant ID in the full info instead. Last year we also had custom URLs for applicants (`admin/assessment/<applicant-id>` to automatically jump to the applicant via a URL). But that can come later if we have time. Is there some other reason you want `id` displayed?

web/containers/admin/AssessmentPanel/index.js
chamkank: Good point, fixed

web/containers/admin/AssessmentPanel/index.js
chamkank: Thanks for pointing this out, fixed the issue by just showing the `ApplicantList` (width 100%) and then showing the score panel and applicant info panel only if some applicant is selected.

web/components/admin/ApplicantList/index.js
chamkank: Yikes looks like I made it a flexbox for no reason

top 10 web developer problems of 2018

web/containers/admin/AssessmentPanel/index.js
chamkank: You're right, `firestore.get` doesn't do any of the automatic stuff that `firestoreConnect` does. 

I went with the second approach you suggested, added `ApplicantInfoPanel` which accepts the `applicantId` and `applicantType` and checks the store / adds listeners if needed for `longInfo` and `shortInfo` for that applicant.

web/containers/admin/AssessmentPanel/index.js
chamkank: Looks like switching between applicant types is bugged now 😢, will fix soon

web/components/admin/ApplicantList/index.js
mingyokim: i'm just confused because applicant 1 is the first applicant displayed and the same applicant 1 can be different based on filters right?

chamkank: That's true, I can't tell from the mocks if that's the intended behavior nor do I remember if that's how it worked last year. But the current functionality might be useful if you want to say `"_______, can you mark resumes for UW applicants 100-200"` or something like that. I'll ask on figma

web/containers/admin/AssessmentPanel/index.js
chamkank: Update: Fixed, also added buttons to switch between the collections.

web/containers/admin/AssessmentPage/index.js
```diff
@@ -0,0 +1,110 @@
...
+  render() {
+    const {
+      applicantType,
+      selectedApplicantId,
+    } = this.state;
+
+    const {
+      firestore,
+    } = this.props;
+    const applicants = firestore.data[applicantCollections[applicantType].shortInfo];
```
mingyokim: shouldn't it be `ordered` instead of `data`?

web/components/admin/ApplicantInfo/index.js
```diff
@@ -0,0 +1,65 @@
...
+  ];
+};
+
+class ApplicantInfo extends React.Component {
+  getApplicantShortInfo = () => {
+    const { firestore, applicantId, applicantType } = this.props;
+    if (!applicantId || !applicantType) {
+      return null;
+    }
+    const shortInfo = firestore.ordered[applicantCollections[applicantType].shortInfo];
```
mingyokim: Why are we using `ordered` here? I think we should use `data` here since we're retrieving a single document. `ordered` is only useful for retrieving a list in order, and takes `O(n)` time to get a document with the right id

web/components/admin/Toolbar/index.js
```diff
@@ -0,0 +1,28 @@
...
+import PropTypes from 'prop-types';
+
+class Toolbar extends React.Component {
+  render() {
+    const { applicantType, switchApplicantType, className } = this.props;
+    return (
+      <div className={`toolbar pad-sides-s pad-top-s fill-width ${className}`}>
+        <i>TODO: Toolbar component</i> <br />
+        Collection: {applicantType}
+        <button type="button" onClick={() => switchApplicantType('hacker')}>Hacker</button>
```
mingyokim: for future reference, I think it would be nice to use something like this for switching applicant type: https://evergreen.segment.com/components/segmented-control

web/components/admin/ApplicantList/index.js
```diff
@@ -0,0 +1,49 @@
...
+      <div className={`applicant-list ${className}`}>
+        {applicants.map((applicant) => {
+          const { id, firstName, lastName, email, finalScore } = applicant;
+          applicantNumber += 1;
+          return (
+            <div
+              className={`applicant flex ${selectedApplicantId === id ? 'selected' : ''}`}
+              key={id}
+              data-applicant-id={id}
+              onClick={onApplicantClick}
```
mingyokim: What do you think about passing in ID in the `onClick` function instead of relying on `data-applicant-id`?

i.e. `onClick={() => onApplicantClick(id)}`

I think especially since we don't have unit testing anymore, it's not really obvious that `data-applicant-id` is a necessary attribute for `onApplicantClick` to work.

web/components/admin/ApplicantInfo/index.js
chamkank: Oops, fixed

web/components/admin/Toolbar/index.js
chamkank: I used the buttons for testing switching between applicants haha, it's supposed to be a select field according to the mocks. I've changed it to use our `<Select>` component and styled it to match the mocks.

Also I'm totally for using third-party components if they're well-maintained and don't look out of place in our current UI (or they do look out of place, but make up for it by providing extra functionality that we currently don't have / can't implement in time). I get the feeling that we'll end up using some for the filter panel.

web/components/admin/ApplicantInfo/index.js
```diff
@@ -0,0 +1,64 @@
...
+  ];
+};
+
+class ApplicantInfo extends React.Component {
+  getApplicantShortInfo = () => {
+    const { firestore, applicantId, applicantType } = this.props;
+    if (!applicantId || !applicantType) {
+      return null;
+    }
+    const shortInfo = firestore.data[applicantCollections[applicantType].shortInfo][applicantId];
```
mingyokim: hmm that's a lot of nested square brackets, i'd suggest breaking this down into multiple lines.

web/components/admin/ScorePanel/index.js
```diff
@@ -0,0 +1,52 @@
...
+  ];
+};
+
+class ScorePanel extends React.Component {
+  getApplicantShortInfo = () => {
+    const { firestore, applicantId, applicantType } = this.props;
+    if (!applicantId || !applicantType) {
+      return null;
+    }
+    const shortInfo = firestore.data[applicantCollections[applicantType].shortInfo][applicantId];
```
mingyokim: same here, i would suggest breaking this down into multiple lines

web/components/admin/Toolbar/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+  render() {
+    const { applicantType, switchApplicantType, className } = this.props;
+    return (
+      <div className={`toolbar fill-width ${className}`}>
+        <Select
+          onChange={option => switchApplicantType(option.value)}
+          label="Form:"
+          name="collection-input"
+          value={this.collectionSelectLabels[applicantType].label}
```
mingyokim: I realized this functionality isn't working properly. If you select an option other than hacker, the dropdown doesn't highlight the current option.

This is one fix:
image

I'm ok if you address this in another pr

web/containers/admin/AssessmentPage/index.js
```diff
@@ -0,0 +1,109 @@
...
+  render() {
+    const {
+      applicantType,
+      selectedApplicantId,
+    } = this.state;
+
+    const {
+      firestore,
+    } = this.props;
+    const applicants = firestore.ordered[applicantCollections[applicantType].shortInfo];
```
mingyokim: nested square brackets - break this into multiple lines?

web/components/admin/Toolbar/index.js
chamkank: What browser are you using? It's working on my end in Chrome and Firefox, value is passed as a string. 

Or are you suggesting to do the object check just in case an object is passed into value?

mingyokim: i'm using chrome. btw it still works, just that when "volunteer" is selected, when you open the dropdown, "hacker" is highlighted. 

I'm suggesting we pass in an object as the `value` in `Select`, and modify the `Select` component to allow objects to be passed in to `value`

chamkank: Ohhh sorry I thought you were talking about the placeholder text, I'll fix this

chamkank:
UGH I forgot to `make build`, was so excited to test this in prod 😞

PR #189, opened 2018-11-20
Add volunteer and mentor application buttons to homepage

christinechengubc:
:tickets: **Ticket(s)**: Closes #187 

---

## :construction_worker: Changes

Add volunteer and mentor application buttons to homepage

## :thought_balloon: Notes

Don't think this is ready to be merged yet. The link to the volunteer applications is still blank as it doesn't exist yet. The link to the mentor application is currently leading to an "edit" page where the mentor application form can be edited. Not sure if the latter issue will be fixed automatically when the mentor application is released.

Also I'm not sure what the branch naming scheme is, but I tried to copy the style of branches that already exist. Furthermore, also unclear if there is a commit message style either. Sorry and thanks for any advice.

Also let me know if I'm doing anything wrong in my code (or could be done better). I'm new to React and generally a noob at coding still. Again many thanks. First PR woohoo!

## :flashlight: Testing Instructions

Run the website and click the buttons.

PR #198, opened 2018-11-26
add feature flags for mentor and volunteer application

mingyokim:

---

Follow up to #189 - buttons for volunteer and mentor application were added, but they shouldn't be visible yet. This is preventing #195 from being deployed, because if that one is built, then these buttons would be visible. Therefore, I added feature flags to be able to dynamically display these buttons based on when these buttons are available to be shown to the users from the firebase console, and allow subsequent PR's to be merged without impacting the visibility of the buttons. 

## :thought_balloon: Notes

Test all combinations of `application` (hacker application), `mentor_application`, and `volunteer_application` feature flags. You can do this from the firebase console in the `feature_flags` collection. Just set `enabled` to `true`/`false` for appropriate documents. Also test responsiveness.


web/components/home/Home.js
```diff
@@ -152,7 +152,17 @@ class Home extends React.Component {
 
     if (!isFeatureFlagsLoaded) return null;
 
-    const { application: { enabled: isApplicationEnabled } } = featureFlagsData;
+    const {
+      application: {
```
chamkank: Minor: this would be better if it were `hacker_application` with `isHackerApplicationEnabled`

PR #201, opened 2018-11-30
Functions/backfill

mingyokim:

---

## :thought_balloon: Notes


mingyokim:
@chamkank good point!! So in https://github.com/nwplus/nwhacks2019/pull/201/commits/15d6245f35fea53c3ef06095f35f57543d50935d, on the client side I generated a token and sent it in the body to the cloud function. Then on the server side it verifies the token and checks if the uid is in `admins`.

mingyokim:
@chamkank I'm thinking of backfilling both short info and long info at once when #206 is done

PR #207, opened 2018-12-05
Functions/backfill long info

### LINKED ISSUE #206: Backfill long info
Similar to #194, but backfill `hacker_long_info` as well

mingyokim:
:tickets: **Ticket(s)**: Closes #206 

---

Basically just did the same thing as #201, but for `hacker_long_info`. There's lots of duplicate code but this is something that's going to be run once and not maintained in the future, so maintainability was not considered when developing this. 

## :thought_balloon: Notes

You can call the backfill functions by navigating to `admin/tasks`.

Create a new doc in `hacker_full_info` with all the required fields.
  - [x] When you press the backfill button for long info, you are prompted with a confirmation modal
  - [x] After confirming, the button is disabled
  - [x] After some time, there is a display message "Success", and the button is re-enabled
  - [x] Check that the corresponding doc is now present in `hacker_long_info`, with the right id, timestamp, and long answer fields.

You can also test the auth stuff by calling the function directly with an invalid token or a token of a non-admin


StanfordLin:
Where do I specify the token?

I'm getting the following error and I believe it's something to do with the token, I'm logged in as admin on localhost:8080

screen shot 2018-12-05 at 1 37 55 am

chamkank:
@StanfordLin that happened to me before, it's because you aren't emulating the functions. It should work if you run `make dev`

@mingyokim This looks good to me because the logic is identical to your last PR, but I'm getting some strange behavior when testing locally. Back-filling short info throws a 500 internal server error (even on master), and back-filling long info returns 1 even when 0 docs are back-filled (it returns the correct number otherwise). Do you have either of these issues on your end?

StanfordLin:
@chamkank  

Thanks, you might have gotten back-filling long info returning 1 because of the text document I created earlier (which wasn't back-filled)

Also, I'm getting a "Cannot find module 'grpc'" and `POST http://localhost:5000/nwhacks-2019-dev/us-central1/backfillLongInfo 500 (Internal Server Error)` after clicking `Backfill long info`

mingyokim:
@chamkank @StanfordLin 

The issue where it always return always 1 is because my code assumes that the `id` field of a document is always identical to the `id` of the doc. But there's a doc that doesn't adhere to this constraint, which is why it's causing unexpected behaviour.

image

mingyokim:
@chamkank @StanfordLin I fixed the resume link and the id field of the document that was causing errors, it should work now

PR #217, opened 2018-12-11
Admin/sorting

### LINKED ISSUE #216: Sorting in the admin panel
Allow client side sorting for the following fields: name, email, score, timestamp (client-side)

Ciuffi:
:tickets: **Ticket(s)**: Closes #216 

---

Added ability to do client side sorting on requested fields. 

I've also added a reverse button to reverse the list order.

## :thought_balloon: Notes

I've added these feature to the toolbar for now, but I'm having trouble formatting it properly. Help would be appreciated.

Example: 
bad formatting

Go to `/admin/assessment` after logging in and try to sort by different fields.

Try pressing the reverse button and make sure the list reverses.

We might want to test this on a bigger data set, but I'm not sure what the best way to do it is.


web/components/admin/AssessmentPage/index.js
```diff
@@ -53,6 +61,15 @@ AssessmentPage.propTypes = {
   switchApplicantType: PropTypes.func,
   // handler for applicant click
   onApplicantClick: PropTypes.func,
+  // what to sort bye
```
chamkank: Typo in the comment

web/components/admin/Toolbar/index.js
```diff
@@ -10,10 +10,17 @@ class Toolbar extends React.Component {
...
+      timestamp: { value: 'timestamp', label: 'Timestamp' },
+    };
   }
 
   render() {
-    const { applicantType, switchApplicantType, className } = this.props;
+    const { applicantType, switchApplicantType, className,
+      switchSortType, sortType, reverseSort, reverse } = this.props;
     return (
       <div className={`toolbar fill-width ${className}`}>
```
chamkank: When in doubt add `flex` to the `className` of the container, fixes 99% of problems I swear 😛

web/components/admin/Toolbar/index.js
```diff
@@ -22,6 +29,13 @@ class Toolbar extends React.Component {
           name="collection-input"
           value={this.collectionSelectLabels[applicantType]}
           options={Object.values(this.collectionSelectLabels)} />
+        <Select
```
chamkank: Might also want to add `className="margin-left-s"` to this Select component to give it some spacing

web/components/admin/Toolbar/index.js
```diff
@@ -22,6 +29,13 @@ class Toolbar extends React.Component {
...
+          onChange={option => switchSortType(option.value)}
+          label="Sort:"
+          name="sort-input"
+          value={this.sortSelectLabels[sortType]}
+          options={Object.values(this.sortSelectLabels)} />
+        <button onClick={reverseSort}>{!reverse ? 'reverse' : 'unreverse'}</button>
```
chamkank: This works fine but if you want to to look super good, we already have an icon on Figma for ascending/descending order. You can export it and try to replace the button with it (it can just be a regular `` with an `onClick`).
image
To flip it upside down, we have an existing utility class (`.flip-vertical`) in [animations.sass](https://github.com/nwplus/nwhacks2019/blob/master/web/styles/animations.sass#L55-L57) that you can add to the `` conditionally.

web/containers/admin/AssessmentPage/index.js
```diff
@@ -18,6 +18,8 @@ class AssessmentPageContainer extends React.Component {
     this.state = {
       applicantType: 'hacker',
       selectedApplicantId: null,
+      sortType: 'timestamp',
+      reverse: false,
```
chamkank: I'm fine with `reverse` but it could be more descriptive if you called it `sortDirection` and allowed the user to pass in either `"asc"` or `"desc"` as the value. It also makes the default state more clear.

web/containers/admin/AssessmentPage/index.js
```diff
@@ -67,30 +69,88 @@ class AssessmentPageContainer extends React.Component {
...
   }
 
+  switchSortType = (newSortType) => {
+    this.setState({
+      sortType: newSortType,
+      selectedApplicantId: null,
+    });
+  };
+
+  sortApplicants = (sortType, applicants) => {
```
chamkank: This function looks super nice! 👌

web/components/admin/Toolbar/index.js
```diff
@@ -10,10 +10,17 @@ class Toolbar extends React.Component {
       mentor: { value: 'mentor', label: 'Mentor' },
       volunteer: { value: 'volunteer', label: 'Volunteer' },
     };
+    this.sortSelectLabels = {
+      name: { value: 'name', label: 'Name' },
```
chamkank: Do you think we could make this `lastName: { value: 'lastName', label: 'Last Name' },`? As a user it wasn't immediately clear to me that we were using last name to sort. Also feel free to add a `firstName` sort filter if you make this change

web/components/admin/Toolbar/index.js
Ciuffi: Very cool! 

I was wondering where to find those graphics.

web/components/admin/Toolbar/index.js
Ciuffi: Good idea, I'll add in the firstName too

web/containers/admin/AssessmentPage/index.js
```diff
@@ -67,30 +69,88 @@ class AssessmentPageContainer extends React.Component {
...
     const {
       firestore,
     } = this.props;
-
     const shortInfoCollectionName = applicantCollections[applicantType].shortInfo;
     const applicants = firestore.ordered[shortInfoCollectionName];
-
+    let sortedApplicants = this.sortApplicants(sortType, applicants);
+    if (reverse) sortedApplicants = applicants.reverse();
     if (!isLoaded(applicants)) {
```
chamkank: I didn't catch this in my first review but it looks like switching to mentor or volunteer breaks the page because L137 & L138 should not come before this if statement on L139.

web/components/admin/Toolbar/index.js
chamkank: ^ If that last part is a question, you can find it on the Internal page:
image
<details>
<summary>To export, you can click on the icon and then on the bottom of the panel that appears on the right you should be able to export it as an SVG file
</summary>

</details>

Ciuffi: Flip vertical applies an animation when its added, but not when its removed.

Any way to get around this?

chamkank: Oh true, you can fix that by adding a transition to `.sort-arrow`:
```
  .sortArrow
    @include transition(0.2s)
    ....
```
This delay will apply to any value changes of [all animatable properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties), but we're only changing the `transform` property for the arrow so it should be fine.

PR #226, opened 2018-12-19
Web/admins/navbar

mingyokim:

---

* Created a new navbar component for admin panel
* Scrapped routes like `admin/tasks` and `admin/assessment` with a single route `admin/home`, where you can still navigate to different tabs but under the same url
* Redirect to `admin/home` after logging in

## :thought_balloon: Notes

Not sure if having the "backfill" page visible to all admins is a good idea, but probably will be fine?

* Test the routes `admin/tasks`, `admin/assessment` when logged in and when not logged in
* Logging in redirects to `admin/home`
* Logging out still works
* Test that you can navigate between different tabs


web/styles/utils.sass
```diff
@@ -160,3 +160,13 @@
...
+
+.active-tab
+  border-bottom: 4px solid $primary
+
+  @media (hover:hover)
+    &:hover
+      border-bottom-color: $secondary-highlight
+
+.inactive-tab
+  border-bottom: 4px solid white
```
mingyokim: I don't like this approach, let me know if you have a better solution. Basically the text is vertically centered, and the text of the active link has less vertical space because of the bottom border. So I just added the border to all links, and set the colour to white (bg is white too) if they're inactive to make them invisible 🤷‍♂️

chamkank: I think this approach is fine, but maybe we can also combine it into a single `.tab` class with the `inactive-tab` being the default state?

Ex:
```
.tab
  border-bottom: 4px solid white // inactive state by default
  &.active
    border-bottom: 4px solid $primary
  @media (hover:hover)
    &:hover
      border-bottom-color: $secondary-highlight
```

web/components/admin/Navbar/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+  render() {
+    const { pages } = this.props;
+    const { tabIndex } = this.state;
+    return (
+      <div>
+        <nav className="fill-width flex">
+          <div className="flex ai-center jc-start margin-sides-l">
+            <div className="flex ai-center">
+              <img alt="nwHacks" src={logo} />
```
chamkank: Idk how much we care about this, but the nwHacks logo for the internal tools is different in the Figma file (some blue variant). I feel like it better suits the general blue theme that our admin panel has.

chamkank:
> Not sure if having the "backfill" page visible to all admins is a good idea, but probably will be fine?

The current tasks don't do anything if their clicked right (since all docs are backfilled)? Should be fine unless we add something that has an effect if it's clicked multiple times. Or we can add an `isDev` flag on our admin accounts lolol

> Scrapped routes like admin/tasks and admin/assessment with a single route admin/home, where you can still navigate to different tabs but under the same url

I like this change because the `App.js` file looks clean, but we may have to reconsider this in the future if we decide to add features that depend on distinguishing between admin pages via URL. For example, jumping to an applicant on the assessment page via URL (`www.nwhacks.io/#/admin/assessment&id=<user_id>`).

web/components/admin/Navbar/index.js
mingyokim: can we change it if someone brings it up LOL

mingyokim:
@chamkank 

> The current tasks don't do anything if their clicked right (since all docs are backfilled)? Should be fine unless we add something that has an effect if it's clicked multiple times. Or we can add an isDev flag on our admin accounts lolol

Yeah the tasks should be idempotent/shouldn't do anything since they're all backfilled already. I think isDev flag is a bit overkill LOL are you okay if we just leave it as is?

> I like this change because the App.js file looks clean, but we may have to reconsider this in the future if we decide to add features that depend on distinguishing between admin pages via URL. For example, jumping to an applicant on the assessment page via URL (www.nwhacks.io/#/admin/assessment&id=<user_id>).

Yeah that would be nice if we had an item view but we currently don't, so it's a bit awkward to implement in the list view especially with all the filters. And as far as i know that's not a requirement right?

chamkank:
> Yeah the tasks should be idempotent/shouldn't do anything since they're all backfilled already. I think isDev flag is a bit overkill LOL are you okay if we just leave it as is?

Yup no worries, we can add `isDev` (or some other flag per function) if we have any future dev functions that aren't idempotent.

> Yeah that would be nice if we had an item view but we currently don't, so it's a bit awkward to implement in the list view especially with all the filters. And as far as i know that's not a requirement right?

Not a requirement but I can see it being implemented in the future (possibly after nwHacks 2019) as a fun side task. We had it last year and I thought it was pretty handy, future teams may also want a way to hyperlink an applicant. But I'm happy with the changes made in this PR, we can modify it again in the future if we ever get to the task.

mingyokim:
@chamkank 

I changed it so that each tab has its own url (`admin/assessment`, `admin/tasks`, etc). Thought it would be better for our future implementation and `admin/assessment` is the url that others are using right now, so didn't want to introduce more confusions. Mind taking another look?

PR #228, opened 2018-12-22
Web/internal/filter

### LINKED ISSUE #227: Implement applicant filtering feature

chamkank:
:tickets: **Ticket(s)**: Closes #227 

---

This implements the filter modal (all of it except for "filter top n" functionality). It already adds about 800 lines of code so I thought it might be better to push this out first to make the review process less painful.

You can:
- filter by a number of properties, as well as tags
- you can also enter custom fields and values

image

## :thought_balloon: Notes

General implementation:
- After filters are applied, a filter expression is generated (which is a function that takes an applicant as input and returns `true` if the applicant passes the combined filter).
  - This filter expression is applied to a flattened version of each applicant object in `applicants`. This means that if the applicant has some nested value like `"tags": {"someTag": true}`, it will be flattened into `"tags_someTag": "true"`. Note that all primitive values in a flattened applicant are converted into a string, but this is fine because JavaScript does type conversion when applying the filters (with non-strict operators like `==` and `!=`).
- I've commented the functions that I've added to make it easier to digest, let me know if something needs clarifying.

Regrets:
- The UI logic took a while to implement and ended up being a ton of code even after refactoring. Might have been better to push out a simple "Enter filter expression" text box to save time but oh well lol

- Try applying filters. We don't have any applicants with tags yet, but you can manually add tags to some applicant to test it out.
- Check that the Apply button enables/disables at the right time
- Check that the Cancel button reverts any changes to the filters
- Try anything you want to try and break the UI state / filters


web/components/admin/FilterModal/index.js
```diff
@@ -0,0 +1,349 @@
...
+    this.state = {
+      filters: getAppliedFilters(),
+      applyButtonDisabled: true,
+    };
+  }
+
+  componentDidMount() {
+    const { store } = this.context;
+    const { firestore } = store;
+    firestore.get('todos');
```
mingyokim: ???

mingyokim:
1. Don't think `education IS High School` is working, at least "4@gmail.com" should qualify but nothing qualifies right now

2. I couldn't get tag to work, I added a field `tags: { accepted: true }` to a doc in `hacker_short_info` and the filter didn't work when setting `accepted IS true`

web/components/admin/FilterModal/index.js
chamkank: Oof lol, that should be tags

chamkank:
1. Nice catch, looks like high school is stored as "High school" in the database, will fix.
2. Do you have the latest version of this branch? I just tried it locally and `a@a.com` is showing up for me.

chamkank:
Thanks! I'm going to merge this so that designers can use the `isDesigner` filter but feel free to leave comments on the code if you find something that could be done better and I'll open a new PR to address the comments.

PR #231, opened 2018-12-27
Web/fix/finalScoreUpdate

chamkank:

---

Previously, `finalScore` was updated *after* the score for a particular criteria was set. This introduced two possible failure cases:
1. The update of a criteria score succeeds, but the `finalScore` update fails, leaving a fully marked applicant with no final score.
2. The update of a criteria score succeeds, but the local store doesn't refresh in time, meaning that the `finalScore` isn't updated at all.

A couple of applicants in prod were seen with this issue, not sure which failure case caused it. Both failure cases were fixed by making the criteria score update + `finalScore` update an atomic operation (either the updates to both fields succeed, or both updates fail) by making it a single firestore call instead of two chained firestore calls.

- [ ] Check that scoring still works as expected.


chamkank:
Thanks for the review!

PR #234, opened 2018-12-31
web/components/volunteerRSVP

lynnzhang7:

Volunteer RSVP page, same as the hacker one but all data is to/from volunteer_short_info. Changed the checkValidID and submitRSVP cloud functions so that they accept an applicant type, so when testing make sure to check that the hacker RSVP still works as before.

## :thought_balloon: Notes

Might need to update the RSVP content, currently uses the hacker content

Also still getting this error, this popped up while I was implementing the original RSVP page and I never figured out why

screen shot 2018-12-31 at 12 23 00 am

Go to http://localhost:8080/#/volunteer-rsvp/5WvE0ayHicEWM082iF5T (try changing the ID)


lynnzhang7:
@chamkank ready for review!

functions/src/RSVP/checkValidID.js
```diff
@@ -9,8 +9,16 @@ exports.checkValidID = functions.https.onRequest((request, response) => {
   response.set('Access-Control-Allow-Headers', 'Content-Type, crossDomain');
 
   const { id } = JSON.parse(request.body);
+  const { applicantType } = JSON.parse(request.body);
```
chamkank: For future reference, you can store the result of `JSON.parse(request.body)` in some variable and use that instead of parsing the body twice, or use a single destructing assignment `const { id, applicantType} = JSON.parse(request.body)`.

lynnzhang7: Good point, forgot I could do that. Just fixed!

lynnzhang7:
@chamkank No problem and will do! Do you have any idea why the build is failing? I think it's cause of this error but I've never seen it before: 
screen shot 2019-01-06 at 11 28 27 pm

chamkank:
@lynnzhang7 
Hmm that's odd, looks like the node process ran out of memory for `make build`. I pushed a commit to this branch that bumps up the allowed RAM usage, should work after that.

PR #236, opened 2018-12-31
Web/admin/applicants page

### LINKED ISSUE #235: Implement applicants page + tagging UI

chamkank:
:tickets: **Ticket(s)**: Closes #235 

---

Mostly implemented, left some smaller items under 'todo' that will be done in the afternoon. In it's current state you can create tags, select specific groups of applicants, and tag/untag them. Pushed it out so that the reviews are less rushed.

**Implemented:**
- `/#/admin/applicants` page
  - ability to view tags
  - ability to select applicants with checkboxes
  - tagging UI
    - can create new tags
    - can select and tag applicants
    - checkboxes include an "indeterminate" state that's shown when only some of the checked applicant's have a specific tag
- `<ApplicantInfo />`: assessment info / full info view for hacker & volunteer
- Added a tie breaker (by timestamp) when sorting by score

**Todo:**
- [x] Show applicantion submission date in `<ApplicantInfo />`
- [x] Only show tag button when applicants are checked
- [x] Move export emails to new toolbar
- [x] Show # of checked applicants
- [x] Filter: select top N (totally forgot about this lol)

Let me know if I've missed anything!

1. `make dev`
2. Go to `/#/admin/applicants`
3. Try creating new tags
3. Try checking + tagging/untagging applicant(s) without any tags
4. Try checking + tagging/untagging applicant(s) with tags
5. Try checking + tagging/untagging applicants with mixed tags. You should see the indeterminate state for the checkboxes if only *some* of the checked applicants have a specific tag.
6. Try cancelling the tag menu (just press out of the modal) and ensure that tags aren't changed for checked applicants


chamkank:
@lynnzhang7 
Thanks for the feedback and for catching all of those errors! Here's what I've fixed:
- [x] `wasn't able to switch to the assessment tab and back, there was some error related to the toolbar`
- [x] `if you don't select any applicants, open the tagging menu and just press apply, a couple errors are thrown but is fixable by refreshing`
  - fixed this my making the tag button only visible when one more more applicants are checked
- [x] `tested filtering by tags and for the most it works great, the UI looks a bit wonky`
  - looks like I wrote a duplicate `tag` CSS class which messed with the filter modal styling, fixed by making them unique
- [x] `export to csv button doesn't work`
  - fixed, should now export the applicants that are currently checked, and is only visible if some applicants are checked

> not sure what the expected behaviour is if you
> 1) tag some applicants
> 2) filter by that tag
> 3) untag those applicants by the same tag
> I think it'd make sense if untaggings triggered the filter function to run again and remove those applicants, or if you can go back into the filter menu and press "apply" again to remove those applicants. Right now the apply button is still disabled after untagging the applicants, and even if i duplicate the filter and press apply, the applicants still stay there (even though they don't apply to the filter). Basically it seems like once you tag an applicant as something, even if you later untag that applicant the applicant will always show up for the filter for that tag.

- Good points! You mention two issues here:
1. >I think it'd make sense if untaggings triggered the filter function to run again and remove those applicants, or if you can go back into the filter menu and press "apply" again to remove those applicants. 
   - I definitely agree with this, right now it's tedious to delete the filter + apply it, then reapply the previous filter. I tried implementing the filter refresh but I've run into some issues. I'll see if I can implement this in time.

2. >Basically it seems like once you tag an applicant as something, even if you later untag that applicant the applicant will always show up for the filter for that tag.
   - Upon further investigation this looks like a bug! After the tags are applied via the cloud function, the local storage isn't updated for some reason. I've fixed this by fetching applicants directly from the DB, so filtering by tags after applying/removing those tags should work fine now. Let me know if it doesn't!

Here's what I most likely won't fix because of the time constraint:
- `no way to delete tags`
  - This was a part of the mocks but when we delete a tag it should delete it from all applicants and doing that requires a cloud function. I don't think it's super necessary for us to send acceptance emails so I'm going to backlog this. If you need to delete a tag from a specific applicant, there's a small X button next to each tag when you click an applicant.

I also probably won't get to the remaining 2 todo items mentioned in the PR tonight because I just don't have the time, but I will work on them tomorrow. Thanks again for your review!

chamkank:
Sorry for all of the commits, I found the source of the bug that caused the following issue you mentioned:
>Basically it seems like once you tag an applicant as something, even if you later untag that applicant the applicant will always show up for the filter for that tag.

Initially I couldn't find the reason for the bug so I made our `GeneralApplicantContainer` manually fetch from the DB. However after finding the bug, I reverted all of those commits.

The real issue was that we were using `firestore.ordered` (which is an array of applicants) but it seems that updates to the array are very inconsistent. However `firestore.data` doesn't suffer from this problem and updates just fine. 😕 So instead of using `firestore.ordered`, I use `Object.values(firestore.data)`.

**tl;dr** of the past two comments - everything should work now. Filter top N is the only thing left to be implemented.

chamkank:
I've added a range/equality filter for `finalScore` instead of filter top N, because adding filter top N would have made the code more complicated (since it's completely different from the current filter structure `(attribute + field + operator + value)`), 

You can still accomplish filter top N with this new functionality. For example, to select the top 200 applicants (by score), just sort by score, scroll down to the 200th applicant and check what their score is, and apply a new filter that returns only applicants with scores `<=` or `>=` that score. Then just uncheck the applicants that go beyond number 200.

With that, this PR is finalized. Any final reviews would be appreciated! 🙏

lynnzhang7:
All the UI stuff looks good, was able to switch to assessment and back, buttons looked good. I also really quickly tested the range equality stuff and it looks great, although it seems like anything that isn't marked should be given a score of zero (is there another way to show only all the unscored applicants?)

Some bugs I found:
- for some reason I can't see the first information for Allison Two? Like when I click on Allison Two it collapses the tags/short info instead of displaying them (also once I select Allison Two I can't seem to deselect, this only happens for allison two though), there's a couple errors thrown
- if i select all the applicants, and try to tag them, I can't see any of the old tags (there's an error:
screen shot 2019-01-02 at 11 11 34 am
and I can't apply tags to all of them, there might be one or two applicants throwing an error cause I'm able to select some groups and tag them) *Carol Ng might be one of them
- UI for the filters looks great, but some functionality seems to be missing and I'm not sure if it's related to the above, I just tried filtering by sample2 and no applicants pop up even though there are a few applicants with that tag
- also getting this error 
screen shot 2019-01-02 at 11 29 04 am

Thanks again for working so hard on this Cham, this functionality looks so complicated so thanks for taking it on!

chamkank:
@lynnzhang7 
Thanks again for your review!

For points 1, 2, and 4 you can fix them by clearing your cache, both Allison Two and Carol Ng were removed from the dev database a while ago (I removed both because Allison Two was missing an ID and a bunch of other fields, and Carol Ng was copied from the prod database which isn't good because we sometimes post demo gifs/screenshots). Not sure why the data is still persisting on your end.

As for the remaining point 3, thanks for catching that! Looks like I broke all the filters except for score when I implemented the score filter. I looked into it and it was converting all string values into `'NaN'` which ended up making those filters return no results. That's been fixed, let me know if you still have the issue!

chamkank:
Awesome I'm glad it works, might be worth looking into the persistence issue some time in the future.

I'm going to merge this, thanks a bunch for taking the time to review this!!! 🙏

PR #237, opened 2019-01-04
Web/internal/bugfixes

chamkank:
:tickets: **Ticket(s)**: Follow up to #236 

---

Fixes two bugs introduced through #236
1. Page height for assessment page (conflicting CSS class names).
2. If you refresh your cache / try the admin page on incognito, both the assessment and applicants page will throw an error and don't render. It's because in `getAllApplicants()`, `Object.values(...`) is used on a `null` value (in the case where you haven't loaded `hacker_short_info` into cache in the past), which breaks everything. This should be fixed now, but it's hard to test on dev because this case only occurs when you're loading a bunch of data (like on prod) / have poor latency.

To test bugfix 1, just check that the applicants and assessment pages have proper page heights (not too long, not too short). For bugfix 2, you can't really test it until it gets into prod lol.


chamkank:
I'm going to merge this w/o a review, but will address any valid comments with a new PR if need be.

I've also filled the dev DB with 1600 dummy applicants, so the difference in behavior between testing on dev and prod should be really just be latency now.

PR #239, opened 2019-01-09
Web/components/volunteerRSVPupdate

lynnzhang7:

Added a new field in the volunteer RSVP page, asking for available orientation times.
This was done with checkboxes/checkbox group, and then the submitRSVP was updated so that the volunteer RSVP retrieves the extra fields and stores them.

## :thought_balloon: Notes

It'd probably make the most sense if a user isn't able to check both "neither" and one of the other options, I looked into it for a while but couldn't figure out how to trigger a change on another checkbox's value based on a different checkbox.

Right now the orientation times a volunteer is available for are stored as three booleans (jan16orientation, jan17orientation, neitherOrientation) , maybe it would be better to store as a string, ex. "Jan 16 only" "Jan 17 only" "Both" "Neither" "Conflicting Response" (if they check neither + something else)

Check to make sure hacker RSVPs and volunteer RSVPs both work as before, try different values for the orientation availability field.

screen shot 2019-01-08 at 11 42 32 pm


functions/src/RSVP/submitRSVP.js
```diff
@@ -36,6 +28,27 @@ exports.submitRSVP = functions.https.onRequest((request, response) => {
...
+    collection = 'volunteer_short_info';
+
+    // additional fields for volunteer only
+    const {
+      jan16orientation,
+      jan17orientation,
+      neitherOrientation,
+    } = JSON.parse(request.body).data;
+
+    rsvpData.rsvp.orientationAvailability = {
```
chamkank: I like how this is nested 👍

web/components/rsvp/volunteer/index.js
```diff
@@ -138,17 +145,43 @@ class volunteerRSVP extends React.Component {
...
             isSearchable
             allowNewOption
           />
+          <CheckboxGroup
+            name="orientationTimes"
+            label={(
+              <div> In order to volunteer for nwHacks 2019, it is required that you attend one
+                orientation session. Please check all the times you can attend:
+              </div>
+)}>
```
chamkank: I'm surprised that eslint doesn't complain about this haha, can we add a couple of spaces to this line?

chamkank:
>Right now the orientation times a volunteer is available for are stored as three booleans (jan16orientation, jan17orientation, neitherOrientation) , maybe it would be better to store as a string, ex. "Jan 16 only" "Jan 17 only" "Both" "Neither" "Conflicting Response" (if they check neither + something else)

I think the way you're doing it right now is fine! We can just display these values in the `<ApplicantInfo />` component as some checkboxes (that can't be toggled).

web/components/rsvp/volunteer/index.js
lynnzhang7: Yep!

PR #242, opened 2019-01-20
Functions/import mentors

mingyokim:

---

## :thought_balloon: Notes

1. Export the mentor rsvp form as csv from google form
2. Go to the "Storage" tab on "nwhacks-2019-dev" firebase console
3. Upload the csv and copy Download url from the right panel
4. Delete the collection `mentors`
5. Go to `admin/tasks` and paste the download url into text field
6. Click on the import mentors button
7. Check that the mentors have been imported with the corresponding fields


functions/src/importMentors/index.js
```diff
@@ -0,0 +1,84 @@
...
+    xhr.onload = function (event) {
+      const { responseText } = xhr;
+      parse(responseText, (err, output) => {
+        if (err) {
+          return response.status(500).send(err);
+        }
+
+        const batch = db.batch();
+
+        output.forEach((row) => {
```
chamkank: Instead of using the `timestamp` from the CSV, can we use `const timestamp = Date.now()` for every iteration of this loop (for consistency, and to avoid updating existing code which assumes timestamp is in Unix time)? The CSV is ordered by timestamp it seems, so the relative order of the mentors should stay the same.

functions/src/importMentors/index.js
```diff
@@ -0,0 +1,84 @@
...
+          const mentorRef = db.collection('mentors').doc();
+          const mentor = {};
+          row.forEach((column, index) => {
+            const fieldName = fieldNamesOrdered[index];
```
chamkank: Seems like the fieldName is `undefined` for one of the checkbox fields (`I agree to not promote, recruit, or otherwise represent my company at the event.`) which results in an undefined field name: 

image

(tbh we can just not write this field to the DB since it's the same for all applicants, up to you)

PR #243, opened 2019-01-21
Export applicants with arbitrary tags

StanfordLin:

---

Exporting selected applicants should include all fields and tags of each applicant

1. Go to http://localhost:8080/#/admin/applicants
2. Select all applicants, export selected applicants
3. Check the excel spreadsheet, should include all fields and tags of each applicant

screen shot 2019-01-20 at 4 56 15 pm


web/containers/admin/GenericApplicantContainer/index.js
```diff
@@ -460,8 +460,28 @@ class GenericApplicantContainer extends React.Component {
       }
     });
     const filteredApplicants = this.filterApplicants(checkedApplicants);
-    const fields = ['firstName', 'lastName', 'email', 'id', 'tags'];
+    const fields = ['firstName', 'lastName', 'email', 'id', 'tags.accepted', 'tags.abc'];
```
chamkank: Can we set fields to an empty array?

StanfordLin: we could, i think it's better if we keep firstName, lastName, email, and id so it keeps it nicely ordered, what do you think?

chamkank: That's a good point, I'm fine with keeping `firstName`, `lastName`, `email`, and `id`, but can we get rid of `tags.accepted`, `tags.abc`?

PR #244, opened 2019-01-22
Web/components/admin/checkin

lynnzhang7:

Added a text field that executes a fuzzy search on the listed applicants, added dropdown that lists registered NFC devices. When a device is selected and an applicant is clicked, their information will be written to the nfc_devices collection under the appropriate device document.

## :thought_balloon: Notes

I tried to run npm install fuse.js but ran into a bunch of errors, so I ended up just implementing a search on the applicants array using a filter to see if firstName + " " + lastName contains the search text or if the email contains the search text. It's kind of crude but I think this works well enough and it was pretty fast when trying it on all the dev applicants. 

Noticed that selecting mentors doesn't work.

Try searching for applicants using their name or email and using the dropdown to check someone in.


web/containers/admin/GenericApplicantContainer/index.js
```diff
@@ -146,7 +148,17 @@ class GenericApplicantContainer extends React.Component {
       return undefined;
     }
     // this.setState({ applicantIndex: firestore.data[shortInfoCollectionName] });
-    const applicants = Object.values(firestore.data[shortInfoCollectionName]);
+    let applicants = Object.values(firestore.data[shortInfoCollectionName]);
+
+    if (searchText !== '') {
```
chamkank: The search is working great, awesome work!

web/components/admin/Toolbar/index.js
```diff
@@ -67,6 +85,19 @@ class Toolbar extends React.Component {
...
             alt="filter button" />
         </div>
+        <TextInput
+          name="search-applicants"
+          className="pad-left-xxl"
+          placeholder="Search..."
+          onChange={(searchText) => {
+            searchApplicants(searchText);
+          }} />
+        <Select
```
chamkank: About the label misalignment issue that you mentioned in your PR body, the width of the `.toolbar .select-input` is set to `265px` which isn't enough space to hold the label + dropdown. If you change that value to `auto` (in the index.sass file for the Toolbar), it should fix it.

web/components/admin/Toolbar/index.js
```diff
@@ -67,6 +85,19 @@ class Toolbar extends React.Component {
             src={areFiltersApplied ? filterIconSelected : filterIcon}
             alt="filter button" />
         </div>
+        <TextInput
```
chamkank: You should be able to fix the alignment and height of the search box by adding something like the following to the `.toolbar` class (in `index.sass` for the Toolbar)
```
  .text-input
    input
      margin-top: 0
      height: 48px
```

web/components/admin/Toolbar/index.js
```diff
@@ -189,6 +221,8 @@ Toolbar.propTypes = {
   pageType: PropTypes.string,
   // Exports all applicants into a csv
   exportApplicants: PropTypes.func,
+  // fuzzy search on applicant name and email
+  searchApplicants: PropTypes.func,
 };
 
-export default Toolbar;
+export default withFirebase(Toolbar);
```
chamkank: For part B, instead of writing a cloud function, you can use react-redux-firebase to automatically listen to the `nfc_devices` collection. You can do that by replacing the export for the toolbar with `export default compose(firestoreConnect(['nfc_devices']))(Toolbar);`. Then to fetch the data, you can do something like `const { firestore: { data: { nfc_devices : devices } } } = this.props` and then devices will contain the `nfc_devices` documents. 

You can see an example of this in `web\components\admin\FilterModal\index.js` (where it listens to the `tags` collection)

web/containers/admin/GenericApplicantContainer/index.js
lynnzhang7: Thanks Cham!!!

web/components/admin/Toolbar/index.js
lynnzhang7: @chamkank okay so this actually fixed an error I hadn't noticed with the new select input! In my PR message though I meant that the search field/textbox is not vertically aligned, it's like a millimeter to low LOL

screen shot 2019-01-22 at 9 46 21 pm

chamkank: OHHH gotcha, I think the search box can be fixed with https://github.com/nwplus/nwhacks2019/pull/244#discussion_r249742497

lynnzhang7: also I'm trying to make the toolbar more like the mocks, not sure if we'll end up keeping this but this is what the toolbar looks like when i add a search emoji as the label
screen shot 2019-01-22 at 9 52 08 pm

web/components/admin/Toolbar/index.js
lynnzhang7: Worked perfectly thank you!

web/components/admin/Toolbar/index.js
lynnzhang7: SORRY I didn't see your other comment, got the vertical alignment fixed 👍 The emoji search as label didn't work (even with the change and also trying width: auto it looked like the photo above), it was kind of dumb anyway so I'm just going to leave it as is

chamkank: Yeah doesn't seem like it adds any functionality, the search box by itself seems fine!

web/components/admin/Toolbar/index.js
lynnzhang7: Fixed! Had no idea I could do it this way. Thanks!!

PR #240, opened 2019-01-15
Web/internal/rsvp fields

chamkank:
## :construction_worker: Changes

- Fixes sorting on assessment page
- Added the RSVP fields for volunteer and hacker
  - Ideally these should be in separate tabs within `<ApplicantInfo />` but with this PR, I just put them in with the rest of the fields.

## :thought_balloon: Notes
I'm merging this as soon as the status checks pass. These are some pretty simple changes but just wanted to open a PR (instead of committing directly to master) to address any future comments!

PR #246, opened 2019-01-25
Web/fix/internal bugs

chamkank:

1. Remove unused code from my previous commits (#240).
2. Store `id` in mentors to prevent <ApplicantList /> from breaking when viewing mentors (I didn't catch this while reviewing #242).
3. When exporting applicants, use quotes around all values to prevent delimiters in values from shifting the data (ex: `city` has commas which will shift the columns incorrectly).

(1) and (2) are simple changes, to test (3) you can check that exported applicant CSV files are properly formatted (no shifted values).


chamkank:
Thanks Stanford for the quick review!

